PR #8, opened 2024-03-18
add seconds support

dosper7:
I've added cronjob expression support for seconds.

With current version I get this error:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to **NCronJob** will be documented in this file. The project
 
 ### Added
 
--   Ability to set the cron expressions can specify with second-level precision
+-   Ability to set cron expressions with second-level precision
 
 ## [0.10.0] - 2024-03-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to **NCronJob** will be documented in this file. The project
 
 ## [Unreleased]
 
+### Added
+
+-   Ability to set the cron expressions can specify with second-level precision
+
 ## [0.10.0] - 2024-03-18
 
 ### Added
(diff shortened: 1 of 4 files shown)


linkdotnet:
Hey @dosper7 - thanks for providing the PR.
A short question: Do you have a use case at hand where you need these intervals?
Tonight, I will check some of the implications, as I might have to make some changes to them.
Especially to guarantee that jobs are really executed every second.

dosper7:
Yes, for example, I'm using NCrobJob in a service that acts as a "proxy" to fetch metrics to another server because of our network restrictions on the scrapping agents.

linkdotnet:
Valid point - one thing I would like to change is that it should be configurable.
Otherwise it breaks "current" usage - even though there isn't much. The bigger point is that it crontab doesn't allow sub-minute resolution by default.

linkdotnet:
Basically, adding a new property to `NCronJobOptions`, which allows the configuration of seconds. This should be opt-in.

Also, can you add an entry to the CHANGELOG.md? This is used for generating the release notes.

dosper7:
I've added the changes and a unit test as well.

linkdotnet:
Good job - LGTM!

dosper7:
BTW, when do you plan to release it? Do you have any schedule on your mind?

linkdotnet:
There is a [0.11.0-preview](https://www.nuget.org/packages/LinkDotNet.NCronJob/0.11.0-preview) version out.

linkdotnet:
I want to add some more features and things for a final 0.11 non-preview version.

linkdotnet:
Just as a heads-up @dosper7 - I will move the `EnableSecondPrecision` from the registration of services into the registration of the job itself. That has some advantages:
 * Not having to build the container inside the registration method (and being more inline with ASP.NET)
 * Don't enforce second precision for all jobs
 
 This will take effect with a release `>0.12.0`.

PR #21, opened 2024-04-17
Implement Retry Mechanism and Concurrency Control for Jobs

falvarez1:
## Overview
This pull request introduces two major enhancements to NCronJob:

1. Retry Mechanism: To improve the robustness of job executions, especially in cases of transient failures.
2. Supports Concurrency Attribute: To provide fine-grained control over the concurrency behavior of individual jobs.

### Details
1. Retry Mechanism

- **Purpose**: Increases reliability by automatically retrying jobs that fail due to transient issues.
- **Implementation**: Implemented using the Polly library with an exponential backoff strategy, allowing recoverable failures to be retried.
- **Configuration**: Customizable per job, enabling specific settings for retry counts and backoff intervals.
2. Supports Concurrency Attribute

- **Purpose**: Adds control over job concurrency to prevent issues from multiple simultaneous instances of the same job type.
- **Implementation**: New `SupportsConcurrency` attribute introduced to mark job classes that can run concurrently.
- **Default Behavior**: Jobs are assumed to be non-concurrent by default. Scheduler checks this attribute before starting a new instance.
- **Impact**: Prevents simultaneous execution of non-concurrent jobs, reducing potential conflicts or resource issues.

### Justification

- **Retry Mechanism**: Enhances resilience against common transient failures in distributed environments.
- **Concurrency Control**: Ensures job execution integrity by preventing duplicate runs of critical or resource-intensive tasks or tasks that work on shared resources.


sample/NCronJobSample/TestRetryJob.cs
```diff
@@ -0,0 +1,34 @@
+using LinkDotNet.NCronJob;
+
+namespace NCronJobSample;
+
+#pragma warning disable CA1848
```
linkdotnet, line 5: If many of the settings are annonying we should have a separate editorconfig for the sample project.
I am not a big fan of a palette of `prgram warning disable`s

src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
```diff
@@ -33,32 +44,36 @@ public void RunJob(RegistryEntry run, CancellationToken stoppingToken)
...
         }
         catch (Exception exc) when (exc is not OperationCanceledException or AggregateException)
         {
             // This part is only reached if the synchronous part of the job throws an exception
-            AfterJobCompletionTask(exc);
+            await AfterJobCompletionTask(exc, default);
         }
-
-        void AfterJobCompletionTask(Exception? exc)
+        // This needs to be async otherwise it can deadlock or try to use the disposed scope, maybe it needs to create its own serviceScope
```
linkdotnet, line 75: In my mental model, the completion (and therefore the notification server handlers) are somewhat a part of the Job itself and, therefore, share the same scope.

src/LinkDotNet.NCronJob/Configuration/ConcurrencyConfig.cs
```diff
@@ -0,0 +1,8 @@
+namespace LinkDotNet.NCronJob;
+
+/// <summary>
+/// Defines the configuration settings for managing concurrency within the application.
+/// This record is used to configure how many operations can run in parallel.
+/// </summary>
+internal record ConcurrencyConfig(int MaxDegreeOfParallelism);
```
linkdotnet, line 7: For the sake of simplicity we can totally remove that.
I would argue in >80% of cases you either want to have unlimited concurrency or just one single instance of a job.

src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
```diff
@@ -9,17 +10,27 @@ internal sealed partial class JobExecutor : IDisposable
...
 
-    public JobExecutor(IServiceProvider serviceProvider, ILogger<JobExecutor> logger)
+    public JobExecutor(IServiceProvider serviceProvider,
+        ILogger<JobExecutor> logger,
+        IHostApplicationLifetime lifetime)
     {
         this.serviceProvider = serviceProvider;
         this.logger = logger;
+
+        lifetime.ApplicationStopping.Register(() => this.shutdown?.Cancel());
```
linkdotnet, line 22: Wouldn't be `Dispose` here the better approach?

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
```diff
@@ -1,47 +1,149 @@
...
+                        var task = Task.Run(async () =>
+                        {
+                            try
+                            {
+                                await ExecuteJob(nextJob, stopToken);
+                            }
+                            finally
+                            {
+                                semaphore.Release();
+                                runningJobTypes[nextJob.Type] = false;  // Mark as not running
```
linkdotnet, line 81: This can lead to race conditions. The dictionary is not thread-safe

falvarez1:
> In your current implementation retry is "always-on" without an option to opt-out. In general I would argue making this feature opt-in is the more sensible choice. That is how many of the .NET objects work and what I would expect as well.

I agree with you about the defaults, I did not integrate configuration bindings as this was more of an initial concept. However most orchestration systems default to retries because they assume the system is deterministic and Jobs are idempotent. Since neither of those are ensured here we can opt-in instead.

> That is an interesting take, because my gut feeling would also go the opposite direction. Make every job concurrent by default and opt-out of this feature. And here I lean towards the same argumentation as above, it feels more natural (to me).
Many jobs, like generating reports or even sending e-mails (given not the identical receiver and message) should work in a concurrent setup just fine.

I understand the natural concurrency for tasks like generating reports or sending e-mails, but this is more about risk-mitigation. The default non-concurrent setting is a safer approach in a non-deterministic environment where shared resources or external states are involved. This reduces the risks of race conditions and unintended side effects. Ensuring that concurrency is only enabled after a developer explicitly verifies it helps prevent complex debugging issues, making the system more robust and easier to maintain. Funny enough the reason I added this to this PR was actually because I hit debugging issues while testing the cancellation support and having many of the same type Jobs running in parallel.

sample/NCronJobSample/Program.cs
```diff
@@ -30,9 +41,9 @@
 
 app.UseHttpsRedirection();
 
-app.MapPost("/trigger-instant", (IInstantJobRegistry instantJobRegistry) =>
+app.MapPost("/trigger-instant", async (IInstantJobRegistry instantJobRegistry) =>
     {
-        instantJobRegistry.RunInstantJob<PrintHelloWorldJob>("Hello from instant job!");
+        await instantJobRegistry.RunInstantJob<TestCancellationJob>("Hello from instant job!");
```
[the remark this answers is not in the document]
linkdotnet: I am somewhat in favor of this (rather than the user just ignores pro-actively the `Task` object) as it would result i non-breaking changes.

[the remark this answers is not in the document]
falvarez1: Actually it is a fire and forget, just that this is handled inside the CronRegistry.RunInstantJob<TJob>() method. I moved it there because it didn't make sense to me to wait for the Jobs' results as the nature of Jobs needs to be real asynchronous. If we need to get the results of the Job there are other ways to handle this like with callbacks or your notification approach but we'd have to make it a more durable, maybe use an in-memory Job Run repository.

sample/NCronJobSample/TestRetryJob.cs
falvarez1: Yeah that was me being impatient, trying to get a test job working. I can fix this or we can have a separate editorconfig as you mention.

src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
falvarez1: If a Job is cancelled then the subsequent AfterJobCompletionTask would not get triggered within the Try block. Are you suggesting to remove the call from within the catch? I left that there because that's how you had it, but I don't pass the CancellationToken to avoid the notifications from being cancelled. I agree this needs a change.

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
falvarez1: Noted, I'll make the change.

src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
linkdotnet: Fair if we say `Ctrl+c=SIGINT`.

sample/NCronJobSample/Program.cs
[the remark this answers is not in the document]
linkdotnet: Fair - my suggestion is then not to return `Task.Completed` and make the function `void` again.
There is no difference between awaiting this and having it void

linkdotnet:
> I understand the natural concurrency for tasks like generating reports or sending e-mails, but this is more about risk-mitigation. The default non-concurrent setting is a safer approach in a non-deterministic environment where shared resources or external states are involved. This reduces the risks of race conditions and unintended side effects. Ensuring that concurrency is only enabled after a developer explicitly verifies it helps prevent complex debugging issues, making the system more robust and easier to maintain. Funny enough the reason I added this to this PR was actually because I hit debugging issues while testing the cancellation support and having many of the same type Jobs running in parallel.

That is a fair point. Let's go with your approach and if we see that the usage is way different, we can still flip it in the next major version.

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
falvarez1: Converted to ConcurrentDictionary

falvarez1:
> That is a fair point. Let's go with your approach and if we see that the usage is way different, we can still flip it in the next major version.

If you prefer I can add a global optIn optOut approach. For example:
```c#
services.AddNCronJob(builder =>
{
    // Use this if you assume most jobs are safe to run concurrently and exceptions are explicitly handled
    builder.OptInConcurrency();  
});

services.AddNCronJob(builder =>
{
    // Use this if you assume most jobs should not run concurrently and only specific jobs are safe to handle it
    builder.OptOutConcurrency(); 
});
``` 
or
```c#
services.AddNCronJob(builder =>
{
    ...
}).OptOutConcurrency();
```
But this would have a minor breaking change since AddNCronJob would not return an `IServiceCollection`. But do you really want to give people the option to chain service registrations off of `AddNCronJob`?

Then decorate the Job according to how you set your global:
```c#
[SupportsConcurrency(true)]
public class ConcurrentJob : IJob
{
    // Implementation allowing concurrency
}

[SupportsConcurrency(false)]
public class NonConcurrentJob : IJob
{
    // Implementation disallowing concurrency
}
``` 

So if you use `OptInConcurrency` then all Jobs are assumed to support concurrency and you would have to explicitly add `[SupportsConcurrency(false)]` as a way to switch that off just for that Job, and vice-versa.

linkdotnet:
I do prefer the attribute (without any boolean flag) over the other given approaches.
That gives granularity and a much better API experience.

We can think of configuring this via the configuration builder:
```csharp
Services.AddNCronJob(o => 
{
 o.AddJob<MySampleJob>().WithCronExpression("...").AllowMultipleInstances();
});
```

falvarez1:
> I do prefer the attribute (without any boolean flag) over the other given approaches. That gives granularity and a much better API experience.
> 
> We can think of configuring this via the configuration builder:
> 
> ```cs
> Services.AddNCronJob(o => 
> {
>  o.AddJob<MySampleJob>().WithCronExpression("...").AllowMultipleInstances();
> });
> ```

@linkdotnet The challenge with that approach is that the dev can easily do the following (because you're not making the concurrency support inherently a part of the Job):
```c#
builder.Services.AddNCronJob(o =>
{
    o.AddJob<MySampleJob>().WithCronExpression("*/10 * * * *").AllowMultipleInstances();
    o.AddJob<MySampleJob>().WithCronExpression("*/25 * * * *");
});
``` 

It's the same `MySampleJob` Job type but one describes support for concurrency while the other does not. I think this leads to confusion and having us explain which one takes precedence. Or just asking devs not to do that, which is not an ideal API experience.

sample/NCronJobSample/Program.cs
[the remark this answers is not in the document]
falvarez1: > Fair - my suggestion is then not to return `Task.Completed` and make the function `void` again.
> There is no difference between awaiting this and having it void

You're right, there's no difference to the end user hitting that endpoint, I just thought it was a more idiomatic way of expressing the method. I think historically the best practice was to use void for fire-and-forget, but lately I've seen more discussions about returning Task. I'll change it back to void.

src/LinkDotNet.NCronJob/Configuration/ConcurrencyConfig.cs
falvarez1: The reason I added this was to use it to evolve concurrency specific settings. I figured you can add things like `TaskQueueLimit` to control memory usage and managing backpressure, PriorityQueueFilter, or some criteria to manage urgency influencing order of execution, TaskTimeout, and maybe even some policies for throttling (e.g. to prevent calls to external APIs from exceeding backpressure) and retry behavior.
I can remove it if you still feel that simpler is better. However regarding unlimited concurrency, I do think we still need some type of ceiling to prevent accidental over-utilization or resource exhaustion.

linkdotnet:
> @linkdotnet The challenge with that approach is that the dev can easily do the following (because you're not making the concurrency support inherently a part of the Job):
> 
> ```cs
> builder.Services.AddNCronJob(o =>
> {
>     o.AddJob<MySampleJob>().WithCronExpression("*/10 * * * *").AllowMultipleInstances();
>     o.AddJob<MySampleJob>().WithCronExpression("*/25 * * * *");
> });
> ```
> 
> It's the same `MySampleJob` Job type but one describes support for concurrency while the other does not. I think this leads to confusion and having us explain which one takes precedence. Or just asking devs not to do that, which is not an ideal API experience.

That is fair. Given that we already settled on the attribute, I'd propose utilizing this to control the concurrency level:

```csharp
[SupportsConcurrency(MaxDegreeOfParallelism = 3)]
public class MyJob : IJob
```

That would also remove the need for the `ConcurrencyConfig` as this can be checked once at the beginning of the scheduler.

sample/NCronJobSample/Program.cs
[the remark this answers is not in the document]
linkdotnet: I am/would be all on board if there is a difference between awaiting and not awaiting the task.

falvarez1:
> ```cs
> [SupportsConcurrency(MaxDegreeOfParallelism = 3)]
> public class MyJob : IJob
> ```
> That would also remove the need for the ConcurrencyConfig as this can be checked once at the beginning of the scheduler.


I think you're mixing the global concurrency with the Job specific concurrency  (basically it's 1 job with many triggers). Your suggestion of putting it in the attribute means it is specific only to that job. However the way I had coded the ConcurrencyConfig.MaxDegreeOfParallelism was to manage the concurrency of the Scheduler within the application, not the Job.  

Either way I think properly supporting concurrency for both individual Jobs and globally can be a little challenging, but I think I have something you can test.

linkdotnet:
> I think you're mixing the global concurrency with the Job specific concurrency 

Indeed I did - thanks for the clarification.

src/LinkDotNet.NCronJob/Configuration/SupportsConcurrencyAttribute.cs
```diff
@@ -1,10 +1,21 @@
...
+    /// <summary>
+    /// Gets the maximum number of concurrent instances allowed for the job.
+    /// </summary>
+    public int MaxDegreeOfParallelism { get; }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SupportsConcurrencyAttribute"/> class with the specified maximum degree of parallelism.
+    /// </summary>
+    /// <param name="maxDegreeOfParallelism">The maximum number of concurrent instances that can run.</param>
+    public SupportsConcurrencyAttribute(int maxDegreeOfParallelism) => MaxDegreeOfParallelism = maxDegreeOfParallelism;
```
linkdotnet, line 19: I do get your intention and the code. 
Here are my thoughts, and I want your opinion on this one.

Currently, in your implementation, you check if `SupportsConcurrencyAttribute.MaxDegreeOfParallelism > GlobalConfig.MaxDegreeOfParallelism`, then we get an exception because it would give the theoretical possibility of having more jobs than slots available.

What if we move this into the scheduler/executor?
I am proposing this because now users have to provide a value. But imagine I have an idempotent job where I don't care how many instances run in parallel; still, I have to give an arbitrary value.
In this case I just want to flag the job, but don't care if there are 2, 3 or 5 instances in parallel.

So that would basically map to:
```csharp
[SupportsConcurrency(int.MaxValue)]
public class MyIdempotantJob : IJob
```

Basically the signature and class looks like this:
```csharp
public sealed class SupportsConcurrencyAttribute : Attribute
{
     public int MaxDegreeOfParallelism { get; set; } = int.MaxValue;
}
```

And people would resolve it like this:
```csharp
// I don't care how many instances run
[SupportConcurrency]
public class Job : IJob

// I only want 2 instances
[SupportConcurrency(MaxDegreeOfParallelism = 2]
public class Job : IJob
```

What is your take on this?

src/LinkDotNet.NCronJob/Configuration/ConcurrencyConfig.cs
falvarez1: Renamed this to ConcurrencySettings and updated the comments for better clarity. Now, it explicitly indicates that these settings apply to the scheduler's concurrency (all jobs).

src/LinkDotNet.NCronJob/Configuration/SupportsConcurrencyAttribute.cs
falvarez1: > Currently, in your implementation, you check if `SupportsConcurrencyAttribute.MaxDegreeOfParallelism > GlobalConfig.MaxDegreeOfParallelism`, then we get an exception because it would give the theoretical possibility of having more jobs than slots available.
> 
> What if we move this into the scheduler/executor?

So I put this in the NCronJobOptionBuilder so that it throws immediately before anything gets a chance to be scheduled. However now thinking about it you may not want to keep the entire program from running just because one job is misconfigured. It's probably more reasonable to log a warning and automatically set the `concurrencyAttribute.MaxDegreeOfParallelism = settings.MaxDegreeOfParallelism` for that job.

I agree that some developers should be able to run a job without worrying about degrees of parallelism. We should implement a sensible default for such cases. Not sure if MaxValue is the way to go, maybe just keep it to the processor count and add a default ctor like this: 
```csharp
[AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
public sealed class SupportsConcurrencyAttribute : Attribute
{
    public int MaxDegreeOfParallelism { get; }

    public SupportsConcurrencyAttribute(int maxDegreeOfParallelism) => MaxDegreeOfParallelism = maxDegreeOfParallelism;

    public SupportsConcurrencyAttribute() => MaxDegreeOfParallelism = Environment.ProcessorCount;
}
```
If a developer needs more control, they can still adjust settings as needed. I think this should be clearly documented also. 

We might consider tweaking the priority scheduling algorithm for jobs triggered to run immediately, treating them as high priority tasks, bumping them up the queue. Additionally, a 'force execution' override could be useful for bypassing concurrency limits and ensuring the job runs immediately. Let me know your thoughts.

It also makes sense to integrate some of these functionalities directly into the JobExecutor, especially those that don't require scheduling but still benefit from retries etc.

linkdotnet:
All in all, this PR goes in a really good direction - I really appreciate your work and discussion, @falvarez1 !

My open ToDo's for now:
 - The retry policy might be configurable as well and not set by default. To be consistent it could be an attribute just like `SupportConcurrency` is. So you can annotate a job with `RetryPolicy(RetryCount = 3)`
 - Some cleanup in the sample because it should show only the bare minimum without confusing someone that is new to the library
 - Some unit tests for the new behavior (which I can also do once it is merged).

Also thanks for keeping the API stable so we don't have to do a major version bump. If you have time you can also extend the `CHANGELOG.md`. This file is used for creating the release notes. If not, I can pick this up as well.

src/LinkDotNet.NCronJob/Configuration/SupportsConcurrencyAttribute.cs
linkdotnet: PS: Saw that you used `* 4` so ignore that comment.

falvarez1:
> * The retry policy might be configurable as well and not set by default. To be consistent it could be an attribute just like `SupportConcurrency` is. So you can annotate a job with `RetryPolicy(RetryCount = 3)`

Glad you said that because I already had that in mind to submit once this was approved ;-)  I'll include it in this one instead.

> * Some cleanup in the sample because it should show only the bare minimum without confusing someone that is new to the library

I agree. I don't know the best way to isolate each use case as a separate example without literally creating a separate sample project. I'll clean it up as best I can but open to suggestions.

> * Some unit tests for the new behavior (which I can also do once it is merged).

Certainly, just waiting to finalize the design first. Don't want to keep changing them while it's in flux.


> * Updating the `README.md` to showcase the new capabilities.

I'll give it shot (both README and CHANGELOG).

> Also thanks for keeping the API stable so we don't have to do a major version bump.

No worries. I do have other ideas but they may require changes to the API definition. Can discuss separately.

linkdotnet:
> I agree. I don't know the best way to isolate each use case as a separate example without literally creating a separate sample project. I'll clean it up as best I can but open to suggestions.

I do agree - it might be tricky. I do think we need a separate documentation page.
It is almost impossible to showcase all the things the library can do, but keep it very simple and minimalistic.

> Certainly, just waiting to finalize the design first. Don't want to keep changing them while it's in flux.

That makes absolutely sense - quite frankly I do think we are almost there.

> No worries. I do have other ideas but they may require changes to the API definition. Can discuss separately. 

Keep it coming ;) - anyway thanks again for the hard work!

falvarez1:
- Introduced support for a `RetryPolicy `attribute to enhance job error handling. The implementation is designed to be extensible, allowing for customization of retry policies while providing simple, sensible defaults. All related code has been organized into the _RetryPolicies_ folder, with thorough documentation included.
- Moved the responsibility for retrying failed job executions from the `CronScheduler `to the `JobExecutor`. This change justifies that retries should be managed by the executor, especially to support retries triggered by instant job runs and not just scheduled ones. This adjustment makes retry functionalities more broadly applicable.

- Priority Queue Enhancement:
  - Tweaked the priority queue algorithm to accommodate a new `JobPriority `enum, enhancing our job scheduling capabilities.
  - Currently, all jobs default to `Normal `priority, but the system now supports prioritization adjustments as needed. This change is internal and does not affect the external API, preserving backward compatibility and easing future integrations.
- Added a default parameter value to the `SupportsConcurrency `attribute, simplifying the configuration process for anyone who "just wants concurrency" without overthinking it.

src/LinkDotNet.NCronJob/Configuration/SupportsConcurrencyAttribute.cs
```diff
@@ -17,5 +17,8 @@ public sealed class SupportsConcurrencyAttribute : Attribute
     /// </summary>
     /// <param name="maxDegreeOfParallelism">The maximum number of concurrent instances that can run.</param>
     public SupportsConcurrencyAttribute(int maxDegreeOfParallelism) => MaxDegreeOfParallelism = maxDegreeOfParallelism;
+
+    /// <inheritdoc />
+    public SupportsConcurrencyAttribute() => MaxDegreeOfParallelism = Environment.ProcessorCount;
```
linkdotnet, line 22: I thought I saw in an earlier revision that you set `Environment.ProcessCount * 4` (which I would go for, as the majority of jobs are I/O based and therefore jobs will idle by quite a lot).

src/LinkDotNet.NCronJob/RetryPolicies/ExponentialBackoffPolicyCreator.cs
```diff
@@ -0,0 +1,34 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Polly;
+
+namespace LinkDotNet.NCronJob;
+
+/// <summary>
+/// A policy creator that configures an exponential back-off retry policy.
+/// </summary>
+public partial class ExponentialBackoffPolicyCreator : IPolicyCreator, IInitializablePolicyCreator
```
linkdotnet, line 10: Does that have to be `public`?

src/LinkDotNet.NCronJob/RetryPolicies/FixedIntervalRetryPolicyCreator.cs
```diff
@@ -0,0 +1,34 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Polly;
+
+namespace LinkDotNet.NCronJob;
+
+/// <summary>
+/// A policy creator that configures a fixed interval retry policy.
+/// </summary>
+public partial class FixedIntervalRetryPolicyCreator : IPolicyCreator, IInitializablePolicyCreator
```
linkdotnet, line 10: Does that have to be `public`?

src/LinkDotNet.NCronJob/RetryPolicies/FixedIntervalRetryPolicyCreator.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+#pragma warning disable CS8618
+    private ILogger<FixedIntervalRetryPolicyCreator> logger;
```
linkdotnet, line 13:
```suggestion
    private ILogger<FixedIntervalRetryPolicyCreator> logger = default!;
```

src/LinkDotNet.NCronJob/RetryPolicies/FixedIntervalRetryPolicyCreator.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    /// <inheritdoc />
+    public void Initialize(IServiceProvider serviceProvider) =>
```
linkdotnet, line 16: Does it need the `Initialize` around?

The logger can be crated as ctor dependency?

src/LinkDotNet.NCronJob/RetryPolicies/PolicyCreatorFactory.cs
```diff
@@ -0,0 +1,15 @@
...
+internal class PolicyCreatorFactory
+{
+    private readonly IServiceProvider serviceProvider;
+
+    public PolicyCreatorFactory(IServiceProvider serviceProvider) => this.serviceProvider = serviceProvider;
+
+    public TPolicyCreator Create<TPolicyCreator>() where TPolicyCreator : IPolicyCreator, new()
+    {
+        var creator = new TPolicyCreator();
+        (creator as IInitializablePolicyCreator)?.Initialize(serviceProvider);
```
linkdotnet, line 12: That would also reduce code here

src/LinkDotNet.NCronJob/RetryPolicies/ExponentialBackoffPolicyCreator.cs
falvarez1: No, only if we want to give someone access to  it. Originally had an implementation in the Sample project that used it but I removed it to simplify things.

src/LinkDotNet.NCronJob/RetryPolicies/FixedIntervalRetryPolicyCreator.cs
falvarez1: Same as above. No, only if we want to give users access to it.

src/LinkDotNet.NCronJob/RetryPolicies/FixedIntervalRetryPolicyCreator.cs
falvarez1: That class can't use ctor because it's used in the RetryPolicyAttribute<T> attribute as a generic parameter. That's why I had to add this initialization pattern so I could access the IServiceProvider.

linkdotnet: Fair!

src/LinkDotNet.NCronJob/RetryPolicies/PolicyCreatorFactory.cs
linkdotnet: As above - no

falvarez1: Same as above, can't add ctor to types used in generic parameters of attributes.This means that when a generic attribute is used, its construction needs to be fully "closed". https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/generic-attributes

src/LinkDotNet.NCronJob/Configuration/SupportsConcurrencyAttribute.cs
falvarez1: That's still there in the global concurrency limit. If we also add it here then one Job with max instances would saturate the limit. We could increase both though. This requires some pretty good stress/load testing to start to see I/O limits.

linkdotnet: Ahhh yes - I was sure to see it somewhere ;)

No that is fine. leave it as it is - that was my fault.

src/LinkDotNet.NCronJob/RetryPolicies/PolicyCreatorFactory.cs
falvarez1: I can simplify the code and avoid the generic attributes, but then it won't be easy to extend it and allow users to make their own retry policies if they so choose. It comes down to flexibility on the user side vs simplicity on the library side.

linkdotnet: Nono - I am 100% onboard with your changes!

falvarez1:
@linkdotnet I've managed to get all the tests working and have added a few new ones. However, I still need to write additional tests to cover the concurrency and retries features effectively. To facilitate this, I plan to introduce a `JobExecutionCount` property to track the number of executions for each job. This property will not only be important for verifying that both concurrent executions and retries are handled correctly but will also enhance observability on the execution of jobs.

I have also addressed issues with the newly added test `ThrowIfJobWithDependenciesIsNotRegistered()`. There were incompatibilities introduced by the updates in this PR that affected this test, but these have now been resolved.

On a side note, I encountered some difficulties with the `TimeProvider` integration. Initially, using the `TimeProviderFactory.GetTimeProvider()` method from the `TimeProviderExtensions` library resulted in very inconsistent results. After switching to Microsoft's `FakeTimeProvider`, the issue was resolved, and everything started working as expected.

I aim to finish the tests and will update the PR as soon as these enhancements are implemented.

linkdotnet:
Really good job!

> Initially, using the TimeProviderFactory.GetTimeProvider() method from the TimeProviderExtensions library resulted in very inconsistent results. 

Using [ManualTimeProvider](https://github.com/egil/TimeProviderExtensions/) was a very deliberate choice.
Mainly because of this behavior:

MTP

VS

FTP

It seemed counter-intuitive to me that if you advance multiple times only that the last result will be invoked n-times.
To be fair, we don't care differentiate between those facts given in the picture.

> I aim to finish the tests soon and will update the PR as soon as they are completed.

Awesome! Let me know if I can help you!

tests/NCronJob.Tests/TestHelper.cs
```diff
@@ -36,7 +40,25 @@ protected virtual void Dispose(bool disposing)
...
-        using var timeoutTcs = new CancellationTokenSource(100);
+        using var timeoutTcs = new CancellationTokenSource(TimeSpan.FromSeconds(2));
+        try
+        {
+            await Task.WhenAll(GetCompletionJobs(jobRuns, timeoutTcs.Token));
+            return true;
+        }
+        catch (OperationCanceledException)
+        {
+            return true;
```
linkdotnet, line 51: Shouldn't this be `false` as well - indicating that the CTS kicked in

tests/NCronJob.Tests/JobOptionBuilderTests.cs
```diff
@@ -59,4 +59,43 @@ public void ShouldAddMultipleCronJobsEvenWithoutParameters()
...
+        options[0].CronExpression.ShouldBe("* * * * *");
+        options[0].EnableSecondPrecision.ShouldBeFalse();
+        options[0].Parameter.ShouldBe("foo");
+        options[1].CronExpression.ShouldBe("0 * * * *");
+        options[1].EnableSecondPrecision.ShouldBeFalse();
+        options[1].Parameter.ShouldBe("bar");
+    }
+
+    [Fact]
+    public async Task ShouldHandleConcurrentAccessesGracefully()
```
linkdotnet, line 85: Is this a realistic case?
Normally there is a single thread configuring this once.

src/LinkDotNet.NCronJob/Configuration/JobOptionBuilder.cs
```diff
@@ -7,26 +7,51 @@ namespace LinkDotNet.NCronJob;
 /// </summary>
 public sealed class JobOptionBuilder
 {
-    private readonly List<JobOption> jobOptions = [];
+    private readonly ConcurrentQueue<JobOption> jobOptions = [];
```
linkdotnet, line 10: See last comment

tests/NCronJob.Tests/NCronJobIntegrationTests.cs
```diff
@@ -14,12 +14,16 @@ public sealed class NCronJobIntegrationTests : JobIntegrationBase
...
     public async Task CronJobThatIsScheduledEveryMinuteShouldBeExecuted()
     {
-        var fakeTimer = TimeProviderFactory.GetTimeProvider();
+        var fakeTimer = new FakeTimeProvider();
         ServiceCollection.AddSingleton<TimeProvider>(fakeTimer);
         ServiceCollection.AddNCronJob(n => n.AddJob<SimpleJob>(p => p.WithCronExpression("* * * * *")));
         var provider = CreateServiceProvider();
 
-        await provider.GetRequiredService<IHostedService>().StartAsync(CancellationToken);
+        // there may be more than one IHostedService, we want to specifically start the CronScheduler
```
linkdotnet, line 22: Well in a unit test not really I'd argue

falvarez1: I meant that the library may end up with more than one IHostedService. It's easy to imagine you can have another one to handle background queueing or pre-initialization of anything else you want to run out-of-band. Anyhow it's not necessary at the moment so I'll remove it.

tests/NCronJob.Tests/TestHelper.cs
falvarez1: I'm not sure. Maybe I misunderstood the intent of the method. I thought that either waiting for a Job or a Timeout should result in success because of how it was being used and the name of the method. I could use some help with this.

tests/NCronJob.Tests/JobOptionBuilderTests.cs
falvarez1: I added the test to verify that the Tasks were all being created and I discovered that 20% of the time we were not getting 100 tasks. It made it obvious that it wasn't thread safe. That's why I changed the `JobOptionBuilder.jobOptions` to a ConcurrentQueue<JobOption>. 
>Is this a realistic case?
Normally there is a single thread configuring this once.

Yes currently only one would be created at a time so no, it's not a use case... but I was thinking ahead with features like auto Job registrations ;-) 
Probably a case of premature optimization. It doesn't hurt to keep the test unless we also revert the jobOptions to a List.

src/LinkDotNet.NCronJob/Configuration/JobOptionBuilder.cs
falvarez1: Let me know if you want to keep it as List, as indicated above it's not a current use case. It adds thread-safety but I'm sure at some minor cost. 
I can revert and run some benchmarks at a later time.

falvarez1:
I added a couple tests for retries and cleaned up some code. Can you review the tests and see if they behave correctly. I could use some help here if you get a chance. Thanks

falvarez1:
By the way the tests using Polly are conducted in actual time, I couldn't figure out how to get it to work with FAke Time, then I found this https://github.com/dotnet/extensions/issues/4912

tests/NCronJob.Tests/NCronJobRetryTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        await provider.GetRequiredService<IHostedService>().StartAsync(CancellationToken.None);
+
+        fakeTimer.Advance(TimeSpan.FromMinutes(1));
+
+        var jobFinished = await WaitForJobsOrTimeout(1);
+        jobFinished.ShouldBeTrue();
+
+        // Validate that the job was retried the correct number of times
+        // Fail 3 times = 3 retries + 1 success
+        var attempts = await CommunicationChannel.Reader.ReadAsync(CancellationToken);
```
linkdotnet, line 32: Looks solid

tests/NCronJob.Tests/NCronJobRetryTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        attempts.ShouldBe(4);
+    }
+
+    [Fact]
+    public async Task JobWithCustomPolicyShouldRetryOnFailure()
+    {
+        var fakeTimer = new FakeTimeProvider();
+        ServiceCollection.AddSingleton<TimeProvider>(fakeTimer);
+        // 3 retries custom policy MyCustomPolicyCreator
+        ServiceCollection.AddSingleton<MaxFailuresWrapper>(new MaxFailuresWrapper(3));
```
linkdotnet, line 42: WRong Policy - it is the same as the test above, isn't it?

tests/NCronJob.Tests/TestHelper.cs
linkdotnet: I properly named the function very poorly.

The idea is to check that n tasks are completely successful. Therefore if you have a timeout, something went wrong.
It is a safety net around the `ChannelWriter` so that the CI pipeline fails quite early and doesn't run indefentily.

linkdotnet:
> By the way the tests using Polly are conducted in actual time, I couldn't figure out how to get it to work with fake Time, then I found this [dotnet/extensions#4912](https://github.com/dotnet/extensions/issues/4912)

Interesting - have to check out the code later.

src/LinkDotNet.NCronJob/RetryPolicies/RetryPolicyAttribute.generic.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        var factory = new PolicyCreatorFactory(serviceProvider);
+        var policy = factory.Create<TPolicyCreator>(RetryCount, DelayFactor);
+        return policy;
+    }
+
+    /// <inheritdoc />
+    public RetryPolicyAttribute(int retryCount = 3, double delayFactor = 2)
+    {
+        ArgumentOutOfRangeException.ThrowIfNegativeOrZero(retryCount);
```
linkdotnet, line 43: The same should apply to the `delayFactor`

tests/NCronJob.Tests/TestHelper.cs
linkdotnet: I did checkout your branch locally and reverted the `true` inside the timeout back to `false` and tests started to fail.

linkdotnet: The reason seems to be how the `FakeTimeProvider` operates. Inside the `CronScheduler`:
```csharp
await Task.Delay(delay, timeProvider, stopToken);
```

We never progress from here if the `FakeTimeProvider` is used but we do with the `ManualTimeProvider`.
Given that plus the whole retry-mechanism it seems that the current setup isn't ideal.

linkdotnet: The idea is (because we have Timers in place) to autoadvance the Timer by the `ManualTimeProvider`

src/LinkDotNet.NCronJob/RetryPolicies/RetryPolicyBaseAttribute.cs
```diff
@@ -0,0 +1,34 @@
+using Polly;
+
+namespace LinkDotNet.NCronJob;
+
+/// <summary>
+/// Abstract base class for defining retry policy attributes, encapsulating common parameters like retry count and delay factor.
+/// Derived classes must implement the creation of specific retry policies tailored to operational needs.
+/// </summary>
+public abstract class RetryPolicyBaseAttribute : Attribute
```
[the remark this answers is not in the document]
falvarez1: Good point. I'll make the change.

src/LinkDotNet.NCronJob/RetryPolicies/RetryPolicyAttribute.generic.cs
falvarez1: Good point. Although zero could be valid, would just retry without any waiting period in between.

tests/NCronJob.Tests/TestHelper.cs
falvarez1: I reverted the WaitForJobsOrTimeout to return false on operation cancelled. I fixed all tests, just needed to advance the time prior to reading from the Channel. I added an overload to WaitForJobsOrTimeout that allows us to pass the advancement of time necessary to trigger the Job run.

linkdotnet:
@falvarez1 looks really good. My main two points would be
 * Documentation
 * If possible, squash the commits to something "more meaningful"

I guess besides that we would be go to go (or merge in that case).

linkdotnet:
Really good work on this one and thanks for taking the time.

linkdotnet (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,24 @@ All notable changes to **NCronJob** will be documented in this file. The project
 
 ## [Unreleased]
 
+## [2.1.0] - 2024-04-23
```
linkdotnet, line 9: You can remove the version - this will be added by the GitHub release action

CHANGELOG.md
```diff
@@ -6,6 +6,24 @@ All notable changes to **NCronJob** will be documented in this file. The project
...
+
+### Added
+- **Retry Mechanism:** Improved robustness of job executions, especially in cases of transient failures. Includes exponential backoff and fixed interval strategies.
```
linkdotnet, line 12: Please also add your Github handle - you deserve the credit.
(See for example version 2.0.5)

CHANGELOG.md
```diff
@@ -6,6 +6,24 @@ All notable changes to **NCronJob** will be documented in this file. The project
...
+
+### Changed
+- **Concurrency Management:** Implemented improved global concurrency handling techniques within `CronScheduler`. This change enhances flexibility and scalability in managing job executions.
+- **Async Job Executions:** Job executions are now fully async from the moment the scheduler triggers the job, ensuring better performance and resource utilization.
+
+### Fixed
+- **Test Framework Bugs:** Addressed specific bugs in the testing framework, ensuring that all tests are now passing and provide reliable results.
+
+### Contributors
+- Support for concurrent jobs and retries, as well as overall improvements, implemented by [@falvarez1](https://github.com/falvarez1) in PR [#21](https://github.com/linkdotnet/NCronJob/pull/21).
```
linkdotnet, line 25: Okay - that is also a nice way of doing it

linkdotnet (approved):

falvarez1:
> I'd directly release a preview version for folks to test.

I agree, I think this needs a little more user testing. I also want to note that some other features I plan to contribute build upon this work, so figuring out a clear path forward would be beneficial for everyone. I suggest creating a `develop` branch that serves as a staging area for preview features before they're merged into main. This would allow any new PRs that build upon the preview features to be targeted against the `develop` branch, keeping the main branch stable while allowing more streamlined development of new features. What do you think about this approach?

> There are small ordering things inside the README.md I'd like to move around (Migration not in between features for example) - but nothing that blocks merging.

Would it also help to put a ToC at the top that links to the sections?

linkdotnet:
@falvarez1 I would merge the changes into `main` - or do you have any open points?
Big thanks to you for the awesome work - really appreciated, that you took the time and effort!

PR #22, opened 2024-04-17
Time zone support

falvarez1:
## Overview
This Pull Request introduces timezone support to the CronScheduler component of our system. Previously, job scheduling was based solely on the server's local time, which could lead to discrepancies in job execution times, especially in systems deployed across multiple time zones or serving international users.

### Motivation
The necessity for timezone support arises from the need to schedule and execute jobs according to various regional times, accommodating users or operations in different time zones. This enhancement ensures that NCronJob is robust, flexible, and more aligned with global use cases.

### Changes Made

1. RegistryEntry Modification:
  

- [x] Added a TimeZoneInfo property to RegistryEntry. This allows each job to specify the timezone in which its cron schedule should be evaluated.

2. Cron Expression Evaluation:
- [x] Modified the ScheduleJob method to compute the next run time of a job based on the timezone specified in its RegistryEntry. This utilizes the GetNextOccurrence method from the Cronos library, which supports timezone adjustments.

3. Job Scheduling and Execution:
- [x] Adjusted the job scheduling logic to use DateTimeOffset instead of DateTime, providing better support for time zones.
- [x] Ensured that the calculated next run time considers the job's specified timezone, converting UTC times to the appropriate local times as needed.

### Impact
- Reliability: Jobs can now be scheduled according to their designated timezones, increasing the reliability of time-sensitive job executions.
- Usability: By supporting multiple timezones, the system can cater to a wider range of users and use cases, enhancing its overall usability.


falvarez1:
I will need to make some modifications to this code once #21 is merged.

falvarez1:
I resolved the merge conflicts, modified the README and added tests. I'm not sure how you want the CHANGELOG.md handled, it doesn't have a version for the last updates, just marked as unreleased.

linkdotnet:
Maybe I should have explained that earlier a bit more in detail or create a `CONTRIBUTING.md`.

The idea behind the `CHANGELOG.md` is basically only adding in the **Unreleased** section. When I create a release via a GitHub Action, this information will be taken and put into the release page information and (currently buggy) put into the csproj file so that you can see the changes also on NuGet.

The Action will create the version tag etc, so there is no need to enter/know upfront the version information.
Once it's released the GitHub action will make a new commit where it moves everything from Unreleased to the newly created version section in the `CHANGELOG.md`.

But there is a small detail: The workflow distinguishes between stable releases and previews. If it is a preview, then this process will **not** kick in.

sample/NCronJobSample/Jobs/PrintHelloWorldJob.cs
```diff
@@ -12,13 +12,14 @@ public Task RunAsync(JobExecutionContext context, CancellationToken token)
     {
         ArgumentNullException.ThrowIfNull(context);
 
-        LogMessage(context.Parameter);
+        if (!string.IsNullOrEmpty(context.Parameter as string))
```
linkdotnet, line 15: TBH: I would drop the check as we are sure that it is set.

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
```diff
@@ -109,6 +114,27 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
     }
 
+    /// <summary>
+    /// Extracts the milliseconds from a <see cref="TimeSpan"/> value.
+    /// Makes sure the value is within the supported range for what Task.Delay can handle, which is about 49ish hours.
+    /// </summary>
+    /// <param name="value"></param>
+    /// <param name="milliseconds"></param>
+    /// <returns></returns>
+    private static bool TryGetMilliseconds(TimeSpan value, out uint milliseconds)
```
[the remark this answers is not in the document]
linkdotnet: To be honest, I think I would remove that part and we have to create a follow up ticket. 50 days is "nothing" when it comes to CRON jobs.
What if want a quarterly report!?

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
```diff
@@ -119,11 +145,16 @@ private void ScheduleInitialJobs()
...
     {
-        var utcNow = timeProvider.GetUtcNow().DateTime;
-        var nextRunTime = job.CrontabSchedule?.GetNextOccurrence(utcNow) ?? DateTime.MaxValue;
-        // higher means more priority
-        var priorityValue = (int)job.Priority;
-        jobQueue.Enqueue(job, (nextRunTime, priorityValue));
+        var utcNow = timeProvider.GetUtcNow();
+        var nextRunTime = job.CronExpression!.GetNextOccurrence(utcNow, job.TimeZone);
+
+        if (nextRunTime.HasValue)
```
linkdotnet, line 151: If you have a valid CRON expression - isn't that guaranteed?

I know that Quartz has some crazy stuff where you explicitly can set a year and whatnot, but with "standard"<sup>In the sense of Unix</sup> CRON you should always get a value. Did I miss something?

tests/NCronJob.Tests/JobOptionBuilderTests.cs
```diff
@@ -80,4 +81,80 @@ public void ShouldCreateMultipleJobsWithoutAnd()
...
+        var builder = new JobOptionBuilder();
+        var timeZone = TimeZoneInfo.FindSystemTimeZoneById("Eastern Standard Time");
+        builder.WithCronExpression("0 12 * * *", timeZoneInfo: timeZone);  // Noon every day
+
+        var options = builder.GetJobOptions();
+
+        var baseTime = new DateTime(2024, 1, 1, 11, 0, 0, DateTimeKind.Utc); // This is 6 AM EST on Jan 1, 2024
+        var expectedRunTime = new DateTime(2024, 1, 1, 17, 0, 0, DateTimeKind.Utc); // Noon EST is 5 PM UTC on Jan 1, 2024
+
+        var nextRunTime = CronExpression.Parse(options.Single().CronExpression)
```
linkdotnet, line 120: I am not a big fan of such tests, mainly because the test tests mainly `Cronos`.
1. That binds tests to. `Cronos` as well
2. It doesn't test that our logic in the scheduler does the right thing.

I would probably use our normal test bed and inject the timeprovider into a job that gets executed that can expose the current `DateTimeOffset`.

sample/NCronJobSample/Jobs/PrintHelloWorldJob.cs
falvarez1: Not really, I added the check because the console just kept outputting `Message: [Null]`.
context.Parameter is a nullable object and was not being set in the sample test. As an example to users I didn't want this to confuse them if they were seeing this as well.
I do have a few ideas on how to pass more structured data, but I'll vet it out before submitting anything.

linkdotnet: Okay fair point - would it make sense just to add a parameter in this case?

linkdotnet: Well - there should be one:
```csharp
AddJob<PrintHelloWorldJob>(p => p.WithCronExpression("* * * * *").WithParameter("Hello from NCronJob"))
```

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
```diff
@@ -109,6 +114,27 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+    /// <summary>
+    /// Extracts the milliseconds from a <see cref="TimeSpan"/> value.
+    /// Makes sure the value is within the supported range for what Task.Delay can handle, which is about 49ish hours.
+    /// </summary>
+    /// <param name="value"></param>
+    /// <param name="milliseconds"></param>
+    /// <returns></returns>
+    private static bool TryGetMilliseconds(TimeSpan value, out uint milliseconds)
+    {
+        const UInt32 maxSupportedTimeout = 0xfffffffe;
```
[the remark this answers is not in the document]
falvarez1: This is how Microsoft did it for the PeriodicTimer code in net repository. So I left it that way 😁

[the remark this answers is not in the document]
linkdotnet: Oh fair - leave it

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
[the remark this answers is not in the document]
linkdotnet: Well - we could "chunk" the time parts in 49-dayish chunks and use continuation until we have the "right" amount of time. Seems also hacky.

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
falvarez1: No you're correct, but `GetNextOccurrence(utcNow, job.TimeZone)` returns nullable so just wrote it like that. Could've used null-forgiving operator but it just seemed visually more intuitive to write it like this.

tests/NCronJob.Tests/JobOptionBuilderTests.cs
falvarez1: Good point, I'll see what I could do.

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
linkdotnet: Ah no that is a good check tbh.
Shamelessly stolen from the docs of Cronos:

> The `nextUtc` will contain the next occurrence in UTC, *after the given time*, or `null` value when it is unreachable (for example, Feb 30).

sample/NCronJobSample/Jobs/PrintHelloWorldJob.cs
falvarez1: Yes, but when working with nullable values I always have to assume that there is always a more-than-zero change that the value will be null. I think checking for null is a good defensive programming strategy.

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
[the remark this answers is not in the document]
falvarez1: Haha, that's pretty much what I was thinking. Definitely hacky but would work.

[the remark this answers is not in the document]
linkdotnet: I created a ticket - but basically it is easier to have our own "Task.Delay" that operates in a while loop!
Very good finding!!! Thanks

sample/NCronJobSample/Jobs/PrintHelloWorldJob.cs
linkdotnet: Absolutely - I, personally, just differ here in "is this code production or test/demo code".
If it is the latter and I know for certain that something is set, I will not go defensive.

falvarez1: I got you. For test/demo code you want to keep it as concise as possible.

tests/NCronJob.Tests/JobOptionBuilderTests.cs
falvarez1: This is done. The tests are now acting directly on the Job.

linkdotnet:
There seems to be a deadlock (GitHub runner timed out after 6h)

falvarez1:
> There seems to be a deadlock (GitHub runner timed out after 6h)

Indeed there is, I must've missed running the retry tests. Ironically that's the deadlock I was fixing with #30 
Once #30 is merged I can pull that in and try again, but trying to fix this here may be redundant.

falvarez1:
This is a weird one. It works fine using Visual Studio and Jetbrains IDE testing mechanisms. Also works fine if I target the individual tests that are failing using `dotnet test --filter "FullyQualifiedName=NCronJob.Tests.NCronJobRetryTests.JobShouldFailAfterAllRetries"` for example. If I step through the test everything completes just fine. It only seems to deadlock if I run all of the tests using `dotnet test`, although the test itself is marked as **Passed**. Sounds like a race condition, probably with disposing objects.

linkdotnet:
> This is a weird one. It works fine using Visual Studio and Jetbrains IDE testing mechanisms. Also works fine if I target the individual tests that are failing using `dotnet test --filter "FullyQualifiedName=NCronJob.Tests.NCronJobRetryTests.JobShouldFailAfterAllRetries"` for example. If I step through the test everything completes just fine. It only seems to deadlock if I run all of the tests using `dotnet test`, although the test itself is marked as **Passed**. Sounds like a race condition, probably with disposing objects. This is after I merged #30 locally, meaning it doesn't change the behavior 🤔.

It indeed sounds like `Dispose`. For me `EachJobRunHasItsOwnScope` is deadlocking inside the while-loop.

src/LinkDotNet.NCronJob/NCronJobExtensions.cs
```diff
@@ -32,6 +32,9 @@ public static class NCronJobExtensions
         var builder = new NCronJobOptionBuilder(services, settings);
         options?.Invoke(builder);
 
+        // temporary, need to be replaced with a proper configuration
+        services.AddSingleton<ConcurrencyConfig>((sp) => new ConcurrencyConfig(3));
```
linkdotnet, line 36: By the way is this still valid for this PR?

linkdotnet:
On another side note: GitHub runners are only single threaded (v)CPU's IIRC.
So this could also influence `local VS GH Runner`

src/LinkDotNet.NCronJob/NCronJobExtensions.cs
falvarez1: No, that must've made it in when I did the first merge. I'll remove it.

linkdotnet:
> It indeed sounds like `Dispose`. For me `EachJobRunHasItsOwnScope` is deadlocking inside the while-loop.

The test in question has the following definition:
> WithCronExpression("* * * * *").And.WithCronExpression("* * * * *")

On `main` that leads to two entries in the `jobQueue` - on the future branch it doesn't (only 1 entry).

linkdotnet:
Okay I found a part of the culprit -.- and I am not sure why it did work on `main`:
```csharp
cronJobs = jobs.Where(c => c.CronExpression is not null).ToFrozenSet();
```

As we are using a record, two equal definitions of a CRON job will of course only result in one element in a set.
So probably we want a `ImmutableArray<RegistryEntry>` here.

falvarez1:
Oh wow I just found the same thing and came here... I was chasing a red herring thinking the issue was with Retry tests.

falvarez1:
I'd argue that we don't want two separate but identical Schedule entries for the same Job. If the point is to run that same Job twice the concurrency control will help here. But I don't think we should allow the same schedule to be created more than once per Job per service.

linkdotnet:
Well that depends - if you want to run multiple jobs at the same time with different parameters I do think that is totally fine and valid.

For the sake of simplicity I generalized that - so if you explicitly add the same cron expression twice you can run it twice.

For sure we could also turn it around and have the schedule as the principal and the job as dependent so that there is a 1 to n relation ship between schedule and job.
This would only allow one run in this case.

falvarez1:
> Well that depends - if you want to run multiple jobs at the same time with different parameters I do think that is totally fine and valid.

Yes that is valid, wouldn't storing as a record (as IEquatable) determine that they are two separate records and then add it just fine?

Anyhow I converted to ImmutableArray and the test passes
```chasrp
    private readonly JobExecutor jobExecutor;
    private readonly ILogger<CronRegistry> logger;
    private readonly ImmutableArray<RegistryEntry> cronJobs;

    public CronRegistry(IEnumerable<RegistryEntry> jobs,
        JobExecutor jobExecutor,
        ILogger<CronRegistry> logger)
    {
        this.jobExecutor = jobExecutor;
        this.logger = logger;
        cronJobs = [..jobs.Where(c => c.CronExpression is not null)];
    }
```
If you prefer we can decide if this is the correct approach in a separate Issue.

linkdotnet:
Yeah - let's discuss the semantics around this in a separate issue.

It triggered the tests again - they weren't stuck, but some seemed to fail. I will have to check this the upcoming days and change something in the Actions, because .net9.0 interferes with the output so that we can't see the failing test.

src/LinkDotNet.NCronJob/NCronJobExtensions.cs
falvarez1: this was removed

linkdotnet:
LGTM so far - I guess only a rebase / merge is one open thing.

README.md
```diff
@@ -61,6 +61,7 @@ jobs:
 - [x] Integrated in ASP.NET - Access your DI container like you would in any other service
 - [x] Get notified when a job is done (either successfully or with an error).
 - [x] Retries - If a job fails, it will be retried.
+- [x] The job scheduler supports TimeZones. Defaults to UTC time.
```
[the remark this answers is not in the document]
falvarez1: Sure, this plus the rebase/merge. Is that the last thing pending?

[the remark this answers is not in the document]
linkdotnet: Yes and the usual `CHANGELOG.md` stuff.

I am in the midst of creating the documentation - I will move a big part of the README towards the docs.
But that has nothing to do with this PR. All in all great job!

linkdotnet:
@falvarez1  I took the liberty to update the branch for you.

Issue #23, opened 2024-04-19
Exceptions in jobs are swallowed


skarum:
Hi,

I have been testing this and I like a lot of what I see, but have just spent an embarrassing amount of time figuring out why my job never got invoked even though the log said it was.

As far as I can see the job executor swallows all exceptions and nothing bubbles out to the log. https://github.com/linkdotnet/NCronJob/blob/f8eacb2affb483d1e0bb91cd70f32f49b6de4309/src/LinkDotNet.NCronJob/Execution/JobExecutor.cs#L41

Could this be changed or made configurable? I'm happy to make a PR if you are interested.

linkdotnet:
Hey @skarum,

currently, this is "by design". The current approach has the issue, that, if an exception is thrown in the continuation, we can't catch it. That is why we introduced the `INotificationHandler<TJob>` that passes in the `Exception? exc` if there was one.

The interesting bit for me is why the job was never invoked in the first place. The referenced line/function isn't doing much besides really kickstarting the job.

> Could this be changed

Yes absolutely - there is an open PR (#21 by @falvarez1) after which we can more easily implement such logic (or tackled already a big chunk)

skarum:
I think I know why it initially did not get trigged.

when the job gets resolved from the container `ServiceProvider.GetService` is called and even though it returns null the type is still of IJob.

If https://github.com/linkdotnet/NCronJob/blob/9b3fe21cd3549923d1d4802f65546849d827e60a/src/LinkDotNet.NCronJob/Execution/JobExecutor.cs#L30 was changed to ` if (scope.ServiceProvider.GetRequiredService(run.Type) is not IJob job)` it should throw.

linkdotnet:
Hmmm, yes that wasn't the best decision from me, just to log this information out instead of throwing an exception.

The `GetService` was on purpose (instead of `GetRequiredService` which instantly throws), so that, even if one job is not properly registered, it does not interfere with other runs.

In retrospective I do understand that is difficult to find out. Would you prefer a "real" exception here rather than a simple log-statement?

skarum:
I would prefer a real exception as this is not a transient error, but the result of my app being setup wrong.

linkdotnet:
Fair enough - if you want, you can file a PR with the suggested change (`GetRequiredService` instead of checking and logging).

It should not interfere with the ongoing PR (#21).

linkdotnet:
Thanks - much appreciated!

PR #24, opened 2024-04-19
Use GetRequiredService instead of GetService to resolve jobs

skarum:
As discussed in #23 

Use GetRequiredService to resolve jobs.

Added tests + modified test that broke due to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ All notable changes to **NCronJob** will be documented in this file. The project
 ### Changed
 
 - Implementation of the scheduling. Better performance and closed some memory leaks
-- Throw exception if job cannot be resolved with dependencies from the DI container
+- Throw exception if job cannot be resolved with dependencies from the DI container. Reported and implemented by [@skarum](https://github.com/skarum) in [#23)(https://github.com/linkdotnet/NCronJob/issues/23)
 
 ## [2.0.4] - 2024-04-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to **NCronJob** will be documented in this file. The project
 ### Changed
 
 - Implementation of the scheduling. Better performance and closed some memory leaks
+- Throw exception if job cannot be resolved with dependencies from the DI container
 
 ## [2.0.4] - 2024-04-16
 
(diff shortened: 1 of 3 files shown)


tests/NCronJob.Tests/NCronJobIntegrationTests.cs
```diff
@@ -163,6 +165,36 @@ public async Task NotRegisteredJobShouldNotAbortOtherRuns()
...
+        Assert.Throws<InvalidOperationException>(() =>
+        {
+            using var executor = new JobExecutor(provider, NullLogger<JobExecutor>.Instance);
+            executor.RunJob(new RegistryEntry(typeof(JobWithDependency), new JobExecutionContext(null), null), CancellationToken.None);
+
+        });
+    }
+
+    [Fact]
+    public async Task NotThrowIfJobWithDependenciesRegistered()
```
linkdotnet, line 184: That is basically tested by many other tests indirectly. I would remove that test.

linkdotnet (commented):
LGTM! Thanks for providing a fix and a test.

Can you add an entry to https://github.com/linkdotnet/NCronJob/blob/main/CHANGELOG.md as I use this file for releases.

tests/NCronJob.Tests/NCronJobIntegrationTests.cs
skarum: Good point - it's removed

skarum:
> Can you add an entry to https://github.com/linkdotnet/NCronJob/blob/main/CHANGELOG.md as I use this file for releases.

Done

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to **NCronJob** will be documented in this file. The project
 ### Changed
 
 - Implementation of the scheduling. Better performance and closed some memory leaks
+- Throw exception if job cannot be resolved with dependencies from the DI container
```
linkdotnet, line 12:
```suggestion
- Throw exception if job cannot be resolved with dependencies from the DI container. Reported and implemented by [@skarum](https://github.com/skarum) in [#23)(https://github.com/linkdotnet/NCronJob/issues/23)
```

linkdotnet:
Perfect! Thanks again. I will make a release `2.0.5` after the merge!

PR #28, opened 2024-04-27
fix: Allow more than 50 days advancements in scheduler

### LINKED ISSUE #27: Can't schedule Jobs that are 50 days or more in the future
**Describe the bug**
Given a quarterly running job - the time difference between the current run and the next (4 months) exceeds the limit of `Task.Delay`, which can only "wait" for ~50 days (`unit.MaxValue - 1`).

**Expected behavior**
A Cron Job that is only executed every 4 month or even once a year should work without a problem.

## Technical "solution"
We could, in theory, write our own `Task.Delay`:

```csharp
public static async Task DelayLong(TimeSpan targetTimeSpan, TimeProvider timeProvider, CancellationToken cancellationToken = default)
{
    ArgumentOutOfRangeException.ThrowIfLessThanOrEqual(targetTimeSpan, TimeSpan.Zero);

    var remainingTime = targetTimeSpan;
    while (remainingTime > TimeSpan.Zero)
    {
        var delay = TimeSpan.FromMilliseconds(Math.Min(remainingTime.TotalMilliseconds, uint.MaxValue - 1));
        await Task.Delay(delay, timeProvider, cancellationToken).ConfigureAwait(ConfigureAwaitOptions.ForceYielding);
        remainingTime -= delay;
    }
}
```

linkdotnet:
Fixes #27

@falvarez1 you could save some checks in your ongoing PR (#22) if we merge this.
Wanted to let you know as it might create merge conflicts for you - are you fine with the change?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to **NCronJob** will be documented in this file. The project
 
 ### Fixed
 - **Test Framework Bugs:** Addressed specific bugs in the testing framework, ensuring that all tests are now passing and provide reliable results.
+- CRON jobs that are scheduled more than 50 days in the future did throw an exception.
 
 ### Contributors
 - Support for concurrent jobs and retries, as well as overall improvements, implemented by [@falvarez1](https://github.com/falvarez1) in PR [#21](https://github.com/linkdotnet/NCronJob/pull/21).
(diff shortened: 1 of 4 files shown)


falvarez1:
Yeah that's fine, seems straight forward. I'll rebase and fix conflicts if anything.

linkdotnet:
Awesome - thanks for having a look

PR #29, opened 2024-04-28
Fix dispose

linkdotnet:
Hey @falvarez1 - I made some changes in regards to the disposable structure and potential memory leaks.

Let me know if you find a reason against the change.


falvarez1:
No that's perfect. I planned on introducing that construct as well so that takes care of that.

PR #30, opened 2024-04-29
fixed cancellations during retrying

falvarez1:
* fixed race conditions issue with disposing shutdown CancellationTokenSource before creating a new one.

* added more thorough tests to validate Retry cancellation

### DIFF
--- a/src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
+++ b/src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
@@ -8,35 +8,53 @@ internal sealed partial class JobExecutor : IDisposable
 {
     private readonly IServiceProvider serviceProvider;
     private readonly ILogger<JobExecutor> logger;
-    private readonly RetryHandler retryHandler;
-    private bool isDisposed;
-    private CancellationTokenSource? shutdown;
+    private readonly IRetryHandler retryHandler;
+    private volatile bool isDisposed;
+    private readonly CancellationTokenSource shutdown = new();
 
     public JobExecutor(IServiceProvider serviceProvider,
         ILogger<JobExecutor> logger,
         IHostApplicationLifetime lifetime,
-        RetryHandler retryHandler)
+        IRetryHandler retryHandler)
     {
         this.serviceProvider = serviceProvider;
         this.logger = logger;
         this.retryHandler = retryHandler;
 
-        lifetime.ApplicationStopping.Register(() => shutdown?.Cancel());
+        lifetime.ApplicationStopping.Register(OnApplicationStopping);
+    }
+
+    private void OnApplicationStopping()
+    {
+        // First cancel all running jobs
+        CancelJobs();
+
+        // Then dispose of the cancellation token source to free resources
+        Dispose();
+    }
+
+    public void CancelJobs()
+    {
+        // Safely trigger the cancellation
+        // This will cancel every job that is currently running
+        if (!shutdown.IsCancellationRequested)
+        {
+            shutdown.Cancel();
+        }
     }
 
     public async Task RunJob(RegistryEntry run, CancellationToken stoppingToken)
     {
-        // stoppingToken is never cancelled when the job is triggered outside the BackgroundProcess,
-        // so we need to tie into the IHostApplicationLifetime
-        shutdown?.Dispose();
-        shutdown = CancellationTokenSource.CreateLinkedTokenSource(stoppingToken);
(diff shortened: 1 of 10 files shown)


src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
```diff
@@ -8,36 +8,54 @@ internal sealed partial class JobExecutor : IDisposable
 {
     private readonly IServiceProvider serviceProvider;
     private readonly ILogger<JobExecutor> logger;
-    private readonly RetryHandler retryHandler;
-    private bool isDisposed;
-    private CancellationTokenSource? shutdown;
+    private readonly IRetryHandler retryHandler;
+    private volatile bool isDisposed;
```
linkdotnet, line 12: Good catch - one question.
What would be the advantage over a `volatile` field over declaring it non-volatile and use `Volatile.Read(ref isDiposed);` when reading.
The Dispose method isn't called from multiple threads, so there wouldn't be a need for a barrier.

src/LinkDotNet.NCronJob/RetryPolicies/RetryHandler.cs
```diff
@@ -48,11 +53,15 @@ public async Task ExecuteAsync(Func<CancellationToken, Task> operation, JobExecu
...
             {
                 runContext.Attempts++;
-                return operation(cancellationToken);
-            });
+                return operation(ct);
+            }, cancellationToken).ConfigureAwait(false);
+        }
+        catch (OperationCanceledException)
+        {
+            LogCancellationOperationInJob();
```
linkdotnet, line 64: We don't want to rethrow here?

tests/NCronJob.Tests/MockHostApplicationLifetime.cs
```diff
@@ -1,14 +1,45 @@
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 
 namespace NCronJob.Tests;
 
-public class MockHostApplicationLifetime : IHostApplicationLifetime
+public class MockHostApplicationLifetime : IHostApplicationLifetime, IDisposable
```
linkdotnet, line 6: You can make that `sealed` to save some boiler-plate code (like `GC.SuppressFinalize`)

linkdotnet:
On a side note, I'd like to invite you as a collaborator.
Going forward that is much easier (especially with fetching PR branches etc).
Of course, only if you agree to this.

src/LinkDotNet.NCronJob/RetryPolicies/RetryHandler.cs
falvarez1: In this context, we handle `OperationCanceledException `without rethrowing it because cancellations are expected behavior, not errors. The distinction is important because:

1. Expected Behavior: Cancellations are a normal part of the lifecycle of tasks that support interruption, not unexpected events that need escalation.
2. Control Flow Management: By catching and logging the cancellation, we manage the workflow gracefully, confirming that the cancellation was detected. We'll add additional processing once states are in place (via events or something).
3. Avoiding Unnecessary Error Handling: Not rethrowing prevents upstream code at the `JobExecutor` from having to handle expected cancellations as errors, simplifying the `JobExecutor` responsibilities.

So the approach avoids conflating cancellations with genuine errors, keeping the system's error handling clean and setting up for when we add states.

tests/NCronJob.Tests/MockHostApplicationLifetime.cs
falvarez1: done

src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
falvarez1: `isDisposed` is marked volatile for simplicity and readability. Since `Dispose()` is not called from multiple threads (given that `JobExecutor` is a singleton) the main concern is ensuring that the `isDisposed` state is visible across threads that may concurrently check its status. I used this approach to maintain simplicity and to avoid potential errors from forgetting to use `Volatile.Read` each time the variable is accessed. I agree `Volatile.Read` may be a better option for performance but I'd like to do some profiling (separately from this PR) to see if the perf benefits are significant enough before doing any optimizations.

linkdotnet: Absolutely fair!

linkdotnet (approved):

PR #31, opened 2024-04-29
fix: Issue where records are dropped because of set behavior

linkdotnet:
@falvarez1 that should fix your deadlock in #22


falvarez1:
Perfect. I did the same in the other branch but I'll just merge this into it.

falvarez1 (approved):

linkdotnet:
> Perfect. I did the same in the other branch but I'll just merge this into it.

If you already have them - I'll drop this.

PR #35, opened 2024-05-02
Fix 34

linkdotnet:
This PR somewhat "mitigates" the flaky test, but doesn't really fix the underlying issue. So I am somewhat in favor of not closing the ticket or make a follow-up, but it is annoying if almost 50% of our test runs fail.

I also took the liberty to
1.  Introduce a new analyzer for disposables (which already detected some issues), as it is easy to overlook.
2. Remove some code comments as they were redundant (the code already tells the story - no need for comments)

### DIFF
--- a/src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
+++ b/src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
@@ -30,14 +30,15 @@ internal sealed partial class CronScheduler : BackgroundService
         this.jobExecutor = jobExecutor;
         this.registry = registry;
         this.timeProvider = timeProvider;
-        this.logger = loggerFactory.CreateLogger<CronScheduler>();
-        this.globalConcurrencyLimit = concurrencySettings.MaxDegreeOfParallelism;
-        this.semaphore = new SemaphoreSlim(concurrencySettings.MaxDegreeOfParallelism);
+        logger = loggerFactory.CreateLogger<CronScheduler>();
+        globalConcurrencyLimit = concurrencySettings.MaxDegreeOfParallelism;
+        semaphore = new SemaphoreSlim(concurrencySettings.MaxDegreeOfParallelism);
 
         lifetime.ApplicationStopping.Register(() => this.shutdown?.Cancel());
     }
     protected override async Task ExecuteAsync(CancellationToken stoppingToken)
     {
+        shutdown?.Dispose();
         shutdown = CancellationTokenSource.CreateLinkedTokenSource(stoppingToken);
         var stopToken = shutdown.Token;
         stopToken.Register(LogCancellationRequestedInJob);
@@ -143,9 +144,7 @@ internal sealed partial class CronScheduler : BackgroundService
         }
         finally
         {
-            // successful or not, increment the execution count
-            // Note: need to refactor to use Interlocked.Increment to better support multi-threaded execution
-            entry.JobExecutionCount++;
+            entry.IncrementJobExecutionCount();
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
```diff
@@ -143,9 +144,7 @@ private async Task ExecuteJob(RegistryEntry entry, CancellationToken stoppingTok
         }
         finally
         {
-            // successful or not, increment the execution count
-            // Note: need to refactor to use Interlocked.Increment to better support multi-threaded execution
-            entry.JobExecutionCount++;
+            entry.IncrementJobExecutionCount();
```
falvarez1, line 147: Looks good. 

On a related note, we should add another way to track total execution of all Jobs to enhance testing and observability. This approach isolates the counting logic from job execution, aligning with the SRP. We can abstract that out into its own class, something like 

```csharp
public class TotalJobExecutionCounter
{
    private int _count = 0;

    public void Increment()
    {
        Interlocked.Increment(ref _count);
    }

    public void Decrement()
    {
        Interlocked.Decrement(ref _count);
    }

    public void Reset()
    {
        Interlocked.Exchange(ref _count, 0);
    }

    public int GetCount()
    {
        return _count;
    }
}
```

tests/NCronJob.Tests/NCronJobIntegrationTests.cs
```diff
@@ -200,7 +200,8 @@ public async Task ThrowIfJobWithDependenciesIsNotRegistered()
 
         await Assert.ThrowsAsync<InvalidOperationException>(async () =>
         {
-            using var executor = provider.CreateScope().ServiceProvider.GetRequiredService<JobExecutor>();
+            using var serviceScope = provider.CreateScope();
+            using var executor = serviceScope.ServiceProvider.GetRequiredService<JobExecutor>();
```
falvarez1, line 204: Good catch. Was this contributing to the flakyness?

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
linkdotnet: That is a good idea - not sure if we want to add `internal` hooks/events as well for testing.
It is unfortunate but without it is really hard to test without polling.
We might be able to drop some `Task.Delay` calls because of that.

tests/NCronJob.Tests/NCronJobIntegrationTests.cs
linkdotnet: Unfortunately not at all - it is still flaky. The problem is that as long as you have a somewhat beefy CPU, you will not notice.
Locally I used the "Run until failure" feature for multiple minutes and didn't encounter a single faulty run - two runs onto GitHub actions and I got the first red test.

That indicates at least some race condition.

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
falvarez1: I was also thinking about how to best add a event/hook system, this ties in well with state management. I'm going to create a prototype in the next couple days and present in the open discussion

falvarez1 (commented):
If the tests continue to fail randomly then I also agree to keep this open until we can get this stable.
How can I reproduce these results?

falvarez1:
I was finally able to reproduce it, but it's very rare. My tests pass over 90% of the time no matter if I comment out the Task.Delay or not. There's definitely an issue but I wish I could reproduce it consistently.

linkdotnet:
Only a `Task.Delay(200)` will fix that locally when a debugger is attached.
I do understand the issue - as `shutdown.Cancel();` is inherently async and we can't await.

Issue #38, opened 2024-05-03
Allow Scheduled Instant Jobs


linkdotnet:
Currently, we only allow instant jobs - but there might be a need for you to execute jobs with a delay (like in 5 Minutes) or to a specific point in time (providing a `DateTimeOffSet`).

Therefore I would propose the following additional API:
```diff
public interface IInstantJobRegistry
{
   void RunInstantJob<TJob>(object? parameter = null, CancellationToken token = default)
        where TJob : IJob;

+   void RunScheduledJob<TJob>(TimeSpan delta, object? parameter, = null, CancellationToken token = default);
+   void RunScheduledJob<TJob>(DateTimeOffset absoluteTime, TimeZoneInfo timeZoneInfo, object? parameter, = null, CancellationToken token = default);
```

PR #39, opened 2024-05-03
feat: Added RunScheduled with a Delay

linkdotnet:
This tackles in part #38 - I would wait until #22 is merged for the second proposed API.


falvarez1 (approved):
Looks good to me.

linkdotnet:
I added also the second API call. I figured that it doesn't have any dependency to #22

linkdotnet:
Hmm those (new) tests are not failing locally.

Issue #26, opened 2024-04-26
GitHub Page - Documentation


linkdotnet:
The only documentation is in two places:
 1. `README.md`
 2. xmldoc with examples on the given type so that you get hints while coding
 
**If** the library has some traction, I dedicated documentation with some use-cases and workflows should be created to guide users through the library.
It also allows for a bit more order rather than dumping everything into the readme file.

linkdotnet:
MVP: Extract the contents of the `README.md` in various files with a bit more context and link it from the `README.md`.

PR #40, opened 2024-05-04
docs: Added documentation

linkdotnet:
This is the initial version for the documentation.

Basically it deploys a GitHub page and I will add triggers when a new tag is created/manual dispatch to update the docs.


linkdotnet:
FYI @falvarez1 - I created some docs that will be deployed initially to https://linkdotnet.github.io/NCronJob

Closes #26

PR #42, opened 2024-05-04
docs: Move from README to Docs

linkdotnet:
Merge this once #22 is merged and docs are updated.

From that point onwards the `README.md` should only consist out of the bare minimum.


linkdotnet:
@falvarez1 as a small piece of information:

The docs will automatically be created when a new tag is created (done by the release workflow if a stable release is created not a pre-release) or a manual trigger.
The latest and greatest version is retrieved via `git` and put into the docs.

I moved all the stuff you nicely put into the README to the docs. Let me know if you have second thoughts about that.

PR #45, opened 2024-05-07
refactor: Extract logic

linkdotnet:
I refactored some code and added the docs for the TImeZone-Support


falvarez1:
@linkdotnet I think the refactoring to the CronScheduler broke the integration tests. I can't get them to get past `await provider.GetRequiredService<IHostedService>().StartAsync(CancellationToken);` with the update.

linkdotnet:
Yeah I will fix that. Interesting as the CI/CD was green.
That’s on me

PR #49, opened 2024-05-08
fixed WaitForNextExecution in CronScheduler

falvarez1:
Some previous refactoring or merging led to `utcNow` in `WaitForNextExecution` to use `DateTime` type instead of `DateTimeOffset`. Since everything else is in `DateTimeOffset` this was producing an incorrect delay of always less than TimeSpan.Zero. This affected the integration tests as well, depending on the local time.

This PR fixes this issue.


linkdotnet:
Ahhhhhh darn! Thanks.

linkdotnet (approved):

linkdotnet:
As I really f-ed up the production logic, I'll make a release.

PR #50, opened 2024-05-08
feat: Infer job-type for builder

linkdotnet:
As discussed


falvarez1 (commented):
Much nicer than using reflection. Looks good 👍

Issue #46, opened 2024-05-07
EPIC: Job Dependencies


linkdotnet:
# Motivation
Jobs can have certain dependencies and should run once a job is successfully finished (or faulted). While it is currently possible to achieve that via the `IJobNotificationHandler` interface it is rather cumbersome.

Basically we could allow two versions:

# 1. Dependency of a job independent of parameter - just depending on the job type
```csharp
builder.Services.AddNCronJob(options =>  options.AddCronJob<Job>(...)
           .When(
        success => success.RunJob<SuccessJob>(p => p.WithParameter("foo")),
        failure => failure.RunJob<FailureJob>())
    );
```

So even if we have `AddCronJob<Job>` with 12 different CRON jobs and or instant trigger, we are calling the same jobs for success or failure job.

# 2. The specific job (with parameter)
```csharp
builder.Services.AddNCronJob(options =>  options.AddCronJob<Job>(
 p => p.WithCronJob("0 * * * *").WithParameter("bar")
    .When(
        success => success.RunJob<SuccessJob>(p => p.WithParameter("foo")),
        failure => failure.RunJob<FailureJob>())
    );
```

So a specific JobDefinition will define the successor. Different CRON jobs can have different success and failures.

# General Questions
1. How deep do we want to nest this? Is it enough to define the first layer? So if we go with approach one:
```mermaid
flowchart LR
    Job1 -- Success --- Job2
    Job1 -- Faulted --- Job3
```

Which would be the given example - if a user want to declare success/faulted for `Job2` or `Job3` he would need to call `AddJob<Job2>().When(...)`.
We could also directly offer a recursive builder to do this in one go:
```mermaid
flowchart LR
    Job1 -- Success --- Job2
    Job1 -- Failure --- Job3
    Job2 -- Success --- Job4
    Job2 -- Failure --- Job5
    Job4 -- Success --- Job6
```

While this might be "cool," it clutters up a lot of code.

2. Do we want to pass in the `JobExecutionContext` from `Job1` to `Job2` or `Job3` or pass in a completely new one?
If so: What happens if the user specifies a `JobExecutionContext` and a parameter? Shall we have a new property on top of `JobExecutionContext`?

For the initial feature, I would go with approach 1 and only one level of Dependency. So a Job only knows its direct successor, not a recursive list.

@falvarez1 Would love to have your thoughts on this one.

falvarez1:
[...]

PR #51, opened 2024-05-08
Job Dependency

linkdotnet:
Currently this is more of an extensive draft rather than a PR. Basically, this PR tackles #46 to a greater extent.

# The Core Idea
The idea is to define job dependencies like this:
```csharp
builder.Services.AddNCronJob(n => n
    .AddJob<MyJob>(p =>
        p.WithCronExpression("*/20 * * * * *", timeZoneInfo: TimeZoneInfo.FindSystemTimeZoneById("Pacific Standard Time")))
    .When(success: d => d.RunJob<PrintHelloWorldJob>("Success"), faulted: d => d.RunJob<PrintHelloWorldJob>("Faulted"),)
```

So, one job type can define 0..n dependencies that will run after the job has finished.
We only distinguish between "success" and "faulted".
 * Faulted: The principal job did throw an exception
 * Success: No exception thrown

The newly dependent job will get the output of the principal job (`ParentOutput`) as part of the `JobExecutionContext`. I was thinking of widening that to have the following API:
```diff
public sealed record JobExecutionContext(Type JobType, object? Parameter)
{
 + public JobExecutionContext? ParentContext { get; set; }
}
```

At least that seems less arbitrary in terms of what we do and don't push down to dependent jobs.

In theory the user can define a job graph as such:
```mermaid
flowchart LR
    Job1 -- Success --- Job2
    Job1 -- Failure --- Job3
    Job1 -- Success --- Job4
    Job2 -- Success --- Job5
    Job2 -- Failure --- Job6
```

1. We allow multiple `RunJob<TJob>` inside one `When` clause - chaining multiple jobs and pushing multiple jobs into the queue.
2. Allowing multiple `AddJob` with `When(s => s.RunJob)` to chain them further. 

I do think that is the right balance between over flexibility and a solution that is too simple.

## Refactorings
In the same term, I refactored some of our types. I extracted some of the logic and moved some bits here and there.
Mainly, I introduced a `JobQueue`, which is the sticky part between the `QueueWorker` (formerly `CronScheduler`) and the `JobExecutor`).

# Current open topics
- [ ] The job inside `RunJob` will **not** automatically be registered. Still, the user has to call `AddJob` somewhere. Maybe we want to change that as it seems arbitrary.
- [ ] There are absolutely zero tests at the moment
- [ ] And zero documentation
- [ ] I have to double check if our Concurrency-Model and so on is still doing what it should do


linkdotnet:
@falvarez1 I am more and more inclined to merge that PR.
It basically doesn't only add the feature but also
 * Fixes #52, as we now go over the same central entity in every case (`JobQueue` and `QueueWorker`).
 
Also, I renamed a few things, but I will annotate a few things that might need some polishing.

src/LinkDotNet.NCronJob/Configuration/NCronJobOptionBuilder.cs
```diff
@@ -116,4 +118,61 @@ public NCronJobOptionBuilder<TJob> AddNotificationHandler<TJobNotificationHandle
...
+    public NCronJobOptionBuilder<TJob> When(Action<DependencyBuilder<TJob>>? success = null, Action<DependencyBuilder<TJob>>? faulted = null)
+    {
+        if (success is not null)
+        {
+            var dependencyBuilder = new DependencyBuilder<TJob>(true);
+            success(dependencyBuilder);
+            dependencyBuilder.GetDependentJobOption().ForEach(s =>
+            {
+                Services.AddSingleton(s);
+                Services.TryAddSingleton(s.DependentJobType);
```
linkdotnet, line 137: I do think we don't want to force needless registration and therefore we do it automatically here.

src/LinkDotNet.NCronJob/Scheduler/QueueWorker.cs
```diff
@@ -56,17 +60,30 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
 
         try
         {
-            while (!stopToken.IsCancellationRequested && jobQueue.Count > 0)
+            while (!stopToken.IsCancellationRequested)
```
linkdotnet, line 63: We have to remove the `jobQueue.Count > 0` as there might be 0 CRON jobs, but a user wants to trigger instant jobs.

src/LinkDotNet.NCronJob/Scheduler/QueueWorker.cs
```diff
@@ -56,17 +60,30 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
             {
                 runningTasks.RemoveAll(t => t.IsCompleted || t.IsFaulted || t.IsCanceled);
 
+                if (jobQueue.Count == 0)
+                {
+                    await Task.Delay(100, stoppingToken);
```
linkdotnet, line 69: PS: This should be a semaphore rather than Task.Delay

src/LinkDotNet.NCronJob/Scheduler/QueueWorker.cs
```diff
@@ -94,6 +113,12 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         }
     }
 
+    private void RescheduleJobs(object? sender, EventArgs e)
+    {
+        queueWaiter.Release();
```
linkdotnet, line 118: Is this really the correct way?
Or:
```csharp
    private void RescheduleJobs(object? sender, EventArgs e)
    {
        rescheduleTrigger.Cancel();
        queueWaiter.Release();
    }
```

falvarez1: Cancel is for Jobs that are currently executing. I don't think rescheduling should implicitly cancel a running job. However, I could see from a dashboard/CLI the option to cancel the job if it's currently running and you're trying to reschedule. So perhaps add a parameter called `forceCancel = false` where you cancel (if it's running) if the user chooses to cancel.

linkdotnet: Just so that we understand each other right - the idea is not to cancel any job.

The code is for the following:
Imagine you have a job that runs hourly. But now you want to trigger an instant job (which with this PR goes through the same pipeline). Obvious our current approach was to wait exactly one hour - execute the job - and grab the next one.

What my code does is basically tell the JobQueue: "Hey, you have a new entry that might be eligible to execute before the current one - therefore, abort the current waiting and retry".

So no job gets canceled. We basically just go back a step and re-evaluate.

For sure I will add appropriate tests for that!

linkdotnet: Added a specific test to document that runs will not be cancelled.

linkdotnet:
One thing that bothered me in this PR I’d that we don’t distinguish between a JobDefinition and a JobRun. 
while not crucial now, i do feel more and more pressure to have those two objects

falvarez1:
> One thing that bothered me in this PR I’d that we don’t distinguish between a JobDefinition and a JobRun. 
> while not crucial now, i do feel more and more pressure to have those two objects 

Yes I agree we need a clear definition and needs to be distinct, this is necessary for the observability for dashboard etc.

Both Jobs and a Job Runs need their own definition.

src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
```diff
@@ -41,7 +45,7 @@ public void CancelJobs()
         }
     }
 
-    public async Task RunJob(RegistryEntry run, CancellationToken stoppingToken)
+    public async Task RunJob(JobDefinition run, CancellationToken stoppingToken)
```
falvarez1, line 48: I think this may be a good opportunity to separate out the Job definition from the "Run". This here needs to be the Job Run, which has its own Job Run Id that lives only within the lifetime of the Run/Execution. This job run id is different from the Job's Id so that we can correctly correlate the events and logs. We can do this in the state-event-system branch but it will have to be done at some point.

Additionally, chaining the Jobs may also require a top level _Flow Id_ so that the entire job flow can be traced and observed instead of inferring the flow from the parent Job Run.

src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
```diff
@@ -51,13 +55,13 @@ public async Task RunJob(RegistryEntry run, CancellationToken stoppingToken)
 
         // stoppingToken is never cancelled when the job is triggered outside the BackgroundProcess,
         // so we need to tie into the IHostApplicationLifetime
-        using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(shutdown.Token, stoppingToken);
+        using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(shutdown.Token, stoppingToken, run.CancellationToken);
```
falvarez1, line 58: The use of `CancellationToken` in job definitions could be ambiguous, especially in job chaining scenarios. There are two main cancellation scenarios to consider:

1. **Cancellation of a Single Job**: The `CancellationToken` should only affect the job it's passed to, ensuring targeted control.
2. **Cancellation of the Entire Job Chain**: In cases where there is a need to terminate the entire sequence of chained jobs, the `CancellationToken` should propagate across the entire chain to halt all subsequent jobs effectively.

Perhaps introducing two different cancellation tokens or a cancellation policy that can be specified at the job or job flow level would provide clearer control.

linkdotnet: That might be a good idea.
i do think I will take some off the good parts and split up this PR.

linkdotnet:
I will pick the good parts here and close the PR.
The PR did too much anyway.

PR #53, opened 2024-05-15
Anonymous job feature

falvarez1:
## Motivation

This update introduces a method to quickly and easily configure cron-based jobs directly from `Program.cs`, eliminating the need for separate classes and reducing boilerplate code. This feature facilitates the creation of jobs with minimal setup, enhancing developer experience by allowing the configuration of anonymous jobs directly within the startup file. This is not designed as a replacement but an alternative way to define simple jobs with minimal fuss.

### Code PR Specific Checklist
- [x] My code follows the style guidelines of this project and the AspNetCore coding standards.
- [x] My change requires updates to the documentation.
- [x] I have added a subsection in the _CHANGELOG.md_ under the appropriate release.
- [x] I have added, updated, or removed tests as necessary according to my changes.
- [x] All tests passed.

### Feature Description

The feature simplifies job creation, enabling job definitions with minimal code. It also supports service injection and passing of the JobExecutionContext. Here is an example demonstrating how to define a job in `program.cs` which injects `ILogger` and `TimeProvider` in 5 lines of code:

```csharp
var builder = Host.CreateApplicationBuilder(args);

builder.Services.AddNCronJob((ILogger<Program> logger, TimeProvider timeProvider) =>
    logger.LogInformation("The current date and time is {Time}", timeProvider.GetLocalNow())
    , "*/5 * * * * *");

await builder.Build().RunAsync();
```

This method supports both synchronous and asynchronous jobs, providing a convenient wrapper that encapsulates `.AddNCronJob()` and automatically schedules and executes the job. Additionally, it ensures that NCronJob services are registered only once, allowing multiple anonymous jobs to be added to the `IServiceCollection`. This functionality is built on the standard approach introduced in V2:

```csharp
builder.Services.AddNCronJob(async (JobExecutionContext ctx, ILoggerFactory loggerFactory, CancellationToken ct) =>
    {
        var logger = loggerFactory.CreateLogger("MinimalSample");
        logger.LogInformation("Job run Id: {JobId}", ctx.Id);
        await Task.Delay(2000, ct);
    }, "*/5 * * * * *");
```

The delegate jobs are registered as singleton's but the context is run with a scoped lifetime so it will support injection of any type of service lifetime.

Concurrency and retry policies continue to be supported by decorating the lambda expression with the corresponding attributes. Here's an example using both:

```csharp
builder.Services.AddNCronJob(
    [RetryPolicy(retryCount: 4)]
    [SupportsConcurrency(5)]
    (JobExecutionContext context, ILogger<Program> logger) =>
{
    var attemptCount = context.Attempts;

    if (attemptCount <= 4)
    {
        logger.LogWarning("Simulating failure.");
        throw new InvalidOperationException("Simulated operation failure.");
    }

    logger.LogInformation($"Job ran after {attemptCount} attempts");
}, "*/7 * * * * *");
```


src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
```diff
@@ -80,15 +86,15 @@ private async Task ExecuteJob(JobExecutionContext runContext, IJob job, AsyncSer
...
         }
         catch (Exception exc) when (exc is not OperationCanceledException or AggregateException)
         {
             // This part is only reached if the synchronous part of the job throws an exception
-            await AfterJobCompletionTask(exc, default);
+            await AfterJobCompletionTask(exc, serviceScope, default);
         }
         // This needs to be async otherwise it can deadlock or try to use the disposed scope, maybe it needs to create its own serviceScope
-        async Task AfterJobCompletionTask(Exception? exc, CancellationToken ct)
+        async Task AfterJobCompletionTask(Exception? exc, AsyncServiceScope serviceScope, CancellationToken ct)
```
linkdotnet, line 97: `serviceScope` seems unused

src/LinkDotNet.NCronJob/Configuration/NCronJobOptionBuilder.cs
```diff
@@ -61,21 +62,86 @@ public NCronJobOptionBuilder<T> AddJob<T>(Action<JobOptionBuilder>? options = nu
...
+
+        var jobOption = new JobOption
+        {
+            CronExpression = cronExpression,
+            EnableSecondPrecision = determinedPrecision,
+            TimeZoneInfo = TimeZoneInfo.Utc
+        };
+        var cron = GetCronExpression(jobOption);
+
+        var jobName = GenerateJobName(jobDelegate);
```
linkdotnet, line 103: While not as user-friendly - a `Guid.NewGuid()` would suffice as a name as well, doesn't it? You created a somewhat "friendly" name in regards to the dashboard?

sample/MinimalSample/Program.cs
```diff
@@ -0,0 +1,10 @@
+using LinkDotNet.NCronJob;
+
+var builder = Host.CreateApplicationBuilder(args);
+
+builder.Services.AddJob((ILogger<Program> logger, TimeProvider timeProvider) =>
```
linkdotnet, line 5: Really nice - I thought about the naming and whether or not that clashes with other libraries. The only example I could find was: `Foundatio.Extensions.Hosting.Jobs`. So that should be fine then.

src/LinkDotNet.NCronJob/Configuration/DynamicJobFactory.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        else if (returnType == typeof(void))
+        {
+            var lambda = Expression.Lambda<Action<object[]>>(Expression.Block(call, Expression.Default(typeof(void))), param);
+            var action = lambda.Compile();
+            return objects => { action(objects); return Task.CompletedTask; };
+        }
+        else
+        {
+            throw new InvalidOperationException("The job action must return a Task or void.");
```
linkdotnet, line 46:
```suggestion
            throw new InvalidOperationException("The job action must return a Task or void type.");
```

src/LinkDotNet.NCronJob/Configuration/DynamicJobFactory.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+
+    private static Func<IServiceProvider, object>?[] BuildServiceResolvers(ParameterInfo[] parameters) =>
+        parameters.Select(p =>
```
linkdotnet, line 52: I would prefer the whole lambda as a local function. Mainly because the else branch does quite a lot

src/LinkDotNet.NCronJob/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+/// <summary>
+/// Extensions for the <see cref="IServiceCollection"/> to add cron jobs.
+/// </summary>
+public static partial class ServiceCollectionExtensions
+{
+    /// <summary>
+    /// Adds a job using an anonymous delegate to the service collection that gets executed based on the given cron expression.
+    /// </summary>
+    /// <param name="services">The service collection used to register the services.</param>
+    /// <param name="jobDelegate">The delegate that represents the job to be executed.</param>
```
linkdotnet, line 14: Could start - for sure I would add that the delegate has to have a certain structure (only `void` or `Task`) maybe enriched with a small example that showcases that.

src/LinkDotNet.NCronJob/Configuration/StringExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+    public static string GenerateConsistentShortHash(this string input)
+    {
+        var hashBytes = SHA256.HashData(Encoding.UTF8.GetBytes(input));
+        return Convert.ToBase64String(hashBytes)[..8]
+            .RemoveChar('+')
+            .RemoveChar('/')
+            .RemoveChar('=');
+    }
+
+    public static string RemoveChar(this string input, char charToRemove)
```
linkdotnet, line 17: nitpick:

Wouldn't it make it easier to do that in one pass-through and let `RemoveChar` either allow to pass in `params char[] toRemove` or have no parameter and check for those three inside the method?

My point is that still 3 strings are created where only one would be needed.
If we want to go extreme:
```csharp
public static string RemoveChar(this ReadOnlySpan<char> input)
```

linkdotnet (commented):
Really nice feature! I do like it a lot and goes into the direction of what MS does with MinimalAPI.

src/LinkDotNet.NCronJob/Execution/JobExecutor.cs
falvarez1: Yeah I missed that. The last commit I tried undoing some work on moving the service scope to the job execution context, but then ended up not needing it. I'll clean it up.

src/LinkDotNet.NCronJob/Configuration/NCronJobOptionBuilder.cs
falvarez1: Exactly. This is not only required to target/lookup a specific Job but so it can be easier to read from a dashboard or CLI. There are other nicer techniques to get short unique IDs (like [squids](https://sqids.org/dotnet)) but I didn't want to rely on any external packages. This can always be changed.

falvarez1:
> Really nice feature! I do like it a lot and goes into the direction of what MS does with MinimalAPI.

Yeah that was the inspiration. They're doing the same with Aspire and I thought this a neat, quick way to get started without too much fuss.

src/LinkDotNet.NCronJob/Configuration/StringExtensions.cs
falvarez1: Yes, a single-pass approach might seem simpler from the caller's perspective, but we still need to iterate over each character, there's no performance advantage—and regex is notably slow. Although it would shift away from being a 'String extension' using ReadOnlySpan<char> would be optimal for performance. I’ll make some tweaks to the implementation.

linkdotnet: Sorry what I meant is that you call your function three times - therefore you pass through the loop three times.
I probably would do this in one pass through and check for all of those three chars

src/LinkDotNet.NCronJob/Configuration/NCronJobOptionBuilder.cs
falvarez1: Additionally, the use of consistent job names, as opposed to generating new GUIDs, is important for several reasons:

1. **Job Identification Across Restarts**: Job names must remain constant across process restarts to ensure accurate mapping and tracking of job run history.
2. **Distributed Environments**: Consistent job names are essential in scenarios involving multiple instances of the same process. This consistency aids in managing distributed locking mechanisms effectively.

Because we're not deriving from IJob we can't get the Job type's Name because one doesn't exist. This approach generates the a name from the delegate's signature, so as long as the signature doesn't change the name will be consistent. Using a stable identifier for each job enhances our ability to maintain state and coordination across system restarts and distributed deployments, improving reliability and traceability.

linkdotnet: I do like that - don’t get me wrong I wasn’t pushing towards Guid.NewGuid - more interested in the reasoning!

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
```diff
@@ -78,6 +78,13 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                         runningTasks.Add(task);
                         ScheduleJob(nextJob);
                     }
+                    else
```
linkdotnet, line 81: I will probably remove this with the approach from #51 with a semaphore.

sample/MinimalSample/Program.cs
linkdotnet, line 1: I do like the idea of having a separate project/sample per "use-case" or major feature.
We might want to have a dedicated small README (or enhance the "main" README) to guide new users.

src/LinkDotNet.NCronJob/NCronJobExtensions.cs
```diff
@@ -27,11 +30,13 @@ public static class NCronJobExtensions
     {
         // 4 is just an arbitrary multiplier based on system observed I/O, this could come from Configuration
         var settings = new ConcurrencySettings { MaxDegreeOfParallelism = Environment.ProcessorCount * 4 };
-        services.AddSingleton(settings);
-
         var builder = new NCronJobOptionBuilder(services, settings);
         options?.Invoke(builder);
 
+        if (services.IsServiceRegistered<CronScheduler>())
```
linkdotnet, line 36: Currently, I have two ways of thinking:
1. If we use this approach, this should be one line 1 so to speak, not?
2. Or alternatively: Wouldn't it be easier to make that function idempotent?

Basically, always calling `TryAddSingleton` etc (`AddHostedService` already uses `TryAddSingleton` under the hood).

linkdotnet:
Really looks good. The question would be whether or not we want to merge the refactor PR first or this one.

I guess documentation/README we can also do in a follow up

src/LinkDotNet.NCronJob/NCronJobExtensions.cs
falvarez1: As far as I remember I don't believe calling AddHostedService is idempotent. I'll verify this

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
falvarez1: Ok, I'll also profile it to make sure I don't see those CPU spikes this intended on mitigating.

src/LinkDotNet.NCronJob/Configuration/NCronJobOptionBuilder.cs
```diff
@@ -61,21 +63,90 @@ public NCronJobOptionBuilder<T> AddJob<T>(Action<JobOptionBuilder>? options = nu
         return new NCronJobOptionBuilder<T>(Services, Settings);
     }
 
-    private static CronExpression GetCronExpression(JobOption option)
+    /// <summary>
+    /// Adds a job using an asynchronous anonymous delegate to the service collection that gets executed based on the given cron expression.
+    /// </summary>
+    /// <param name="jobDelegate">The delegate that represents the job to be executed.</param>
+    /// <param name="cronExpression">The cron expression that defines when the job should be executed.</param>
+    public NCronJobOptionBuilder AddJob(Delegate jobDelegate, string cronExpression)
```
linkdotnet, line 71: Just from an API point of view: If we also call that `AddNCronJob` would this be fitting?

My reasoning is that the user has one simple entry point.

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
falvarez1: @linkdotnet I've confirmed the CPU spike remains if I remove the Task.Delay(1). I tried a few other approaches but this seems to be the most effective. The CPU was spiking to 20% at one point while the scheduler was just sitting there waiting.

falvarez1: I also confirmed this is an existing issue on the `main` branch, we can keep this fix on this PR or add separate PR just for this fix. 

To reproduce on main just add Job with a long async delay and a short interval like like this
```csharp
public class PrintHelloWorldJob : IJob
{
    public async Task RunAsync(JobExecutionContext context, CancellationToken token) => await Task.Delay(25000, token);
}

builder.Services.AddNCronJob(n => n
    .AddJob<PrintHelloWorldJob>(p =>
        p.WithCronExpression("*/5 * * * * *", timeZoneInfo: TimeZoneInfo.FindSystemTimeZoneById("Pacific Standard Time")))
);
```

sample/MinimalSample/Program.cs
falvarez1: I think we can add another README right on the sample folder (and rename to samples/examples). This should be a separate issue to identify the various use-cases.

src/LinkDotNet.NCronJob/Configuration/NCronJobOptionBuilder.cs
falvarez1: I think you mean the method in `ServiceCollectionExtensions`. I've updated the name to be more consistent with the entry point.
```csharp
    public static IServiceCollection AddNCronJob(this IServiceCollection services, Delegate jobDelegate, string cronExpression)
    {
        services.AddNCronJob(builder =>
            builder.AddJob(jobDelegate, cronExpression)
        );

        return services;
    }
    ```

File: README.md [hunk 11]
```diff
@@ -58,6 +59,21 @@ look into a more advanced scheduler like `Hangfire` or `Quartz`.
 
 ## Short example
 
+There are two ways of defining a job.
+
+### Minimal Job API
+You can use this library in a simple one-liner:
+```csharp
+builder.Services.AddJob((ILogger<Program> logger, TimeProvider timeProvider) =>
```
linkdotnet, line 67: Note for future self: Has to be updated here as well.

linkdotnet (approved):

src/LinkDotNet.NCronJob/NCronJobExtensions.cs
linkdotnet: By the way, I added the idempotent version. `AddHostedService` does indeed call `TryAddEnumerable` which will not work if the same instance is already registered.

sample/MinimalSample/Program.cs
linkdotnet: (Y) Yeah - not in this PR. Would like to merge it and enhance the docs so I can release a stable version.

src/LinkDotNet.NCronJob/Scheduler/CronScheduler.cs
linkdotnet: Hmm - Understandable. We might use this as our current debouncer. I would argue that even `Task.Delay(50)` should be good enough - anyway `Task.Delay(1)` is somewhat in the range of 1 to 16ms depending on the OS and arch (as they result in a different resolution for the timer).

linkdotnet:
Feel free to merge if there aren't any open topics from your side.

PR #55, opened 2024-05-18
refactor: Overhaul mechanism of job scheduling

linkdotnet:
I moved a sub-part of #51 into this one, so we can incrementally merge stuff.
There was too much going on in the other PR.

While there will be merge conflicts with #53 - I strongly suggest merging this first, as some of the logic can be simplified in the other PR.


falvarez1 (approved):
Looks good

PR #60, opened 2024-05-20
refactor: Remove polling and wait for first task to finish

linkdotnet:
That should be a more stable way than `Task.Delay(1);`

I just started a bunch of `PrintHelloWorld` jobs as test - maybe you have a better testbed @falvarez1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to **NCronJob** will be documented in this file. The project
 
 ## [Unreleased]
 
+### Fixed
+
+- Retry/Concurrency Attributes where ignored
+
 ## [2.4.5] - 2024-05-20
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


falvarez1:
@linkdotnet 

The `await Task.WhenAny(runningTasks);` has an issue that I discovered, which may be more of a fundamental issue with the way Jobs are prioritized. If you create a job with a very long delay and short scheduled interval frequency we get to a point where the runningTasks just contain the long running job, so the entire scheduler now has to wait until that one job completes before Task.WhenAny is released.

Using something like `await Task.WhenAny(runningTasks.Concat([Task.Delay(1000, stopToken)]));` gives it a chance to exit and check again after 1 second but I suspect this will affect the scheduling.

We need to push the same Jobs as the long running Job that are at the top of the PriorityQueue down to allow other jobs to be scheduled. Or check if the execution time is past some grace period and if so dequeue it.

linkdotnet:
That us a fair point and somewhat what I would expect as we just drop new runs if the queue is full.

I don’t think that either Task.Delay(1) or your version (with Concat Task Delay(1000)) would help here as the queue is still full.

I am currently more inclined to put this into the docs.
If you have 10 slots and you allow your extremely long running job to have those 10 slots, then yes it can happen that no other job will run.

linkdotnet:
This PR will now only fix the missing meta-data. I will open separate ones for the other related issues.

PR #61, opened 2024-05-21
fix: Throw exception when missing instant job

linkdotnet:
Instant jobs should directly throw if the underlying job is missing inside the container.

### DIFF
--- a/src/LinkDotNet.NCronJob/Registry/JobRegistry.cs
+++ b/src/LinkDotNet.NCronJob/Registry/JobRegistry.cs
@@ -1,81 +1,21 @@
 using System.Collections.Immutable;
-using Microsoft.Extensions.Logging;
 
 namespace LinkDotNet.NCronJob;
 
-internal sealed partial class JobRegistry
+internal sealed class JobRegistry
 {
-    private readonly JobExecutor jobExecutor;
-    private readonly TimeProvider timeProvider;
-    private readonly ILogger<JobRegistry> logger;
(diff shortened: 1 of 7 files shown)


src/LinkDotNet.NCronJob/Registry/JobRegistry.cs
```diff
@@ -1,81 +1,21 @@
...
-        where TJob : IJob => RunScheduledJob<TJob>(TimeSpan.Zero, parameter, token);
-
-    /// <inheritdoc />
-    public void RunScheduledJob<TJob>(TimeSpan delay, object? parameter = null, CancellationToken token = default)
-    {
-        token.Register(() => LogCancellationRequested(parameter));
-
-        var run = new JobDefinition(typeof(TJob), parameter, null, null);
-
-        _ = Task.Run<Task>(async () =>
```
falvarez1, line 38: Is there no use-case where you would want to trigger a job regardless of the concurrency settings for that job type?

falvarez1 (approved):
Code looks good. Just wanted to point out that there may be a use case for force triggering a job, so long as it supports concurrency, and bypass the queue altogether.

src/LinkDotNet.NCronJob/Registry/JobRegistry.cs
linkdotnet: That is a good enhancement, to be honest.
Right now, it will always go via the queue -> executor, which makes a lot of things easier (and easier to reason about).

I will add this as its own enhancement.

PR #68, opened 2024-05-22
added caching mechanism to policy attributes

falvarez1:
## Pull request description
We use reflection during Job registration to gather the metadata around the Job's attribute decorators. This PR creates a caching strategy so we only ever perform the reflection once. This also allows a `JobDefinition` to automatically set its attribute properties to reduce boilerplate code and improve maintainability. As a side-effect this allows an `InstantJobRegistry` Job to support the retry and concurrency attributes.

### PR meta checklist
- [X] Pull request is targeted at `main` branch for code   
- [X] Pull request is linked to all related issues, if any.

- [X] My code follows the code style of this project and AspNetCore coding guidelines.
- [X] All tests passed.

### DIFF
--- a/src/NCronJob/Registry/JobAttributeCache.cs
+++ b/src/NCronJob/Registry/JobAttributeCache.cs
@@ -7,8 +7,8 @@ internal static class JobAttributeCache
     private static readonly ConcurrentDictionary<object, JobExecutionAttributes> Cache = new();
 
     public static JobExecutionAttributes GetJobExecutionAttributes(Type jobType) =>
-        Cache.GetOrAdd(jobType, (type) => JobExecutionAttributes.CreateFromType(type as Type));
+        Cache.GetOrAdd(jobType, (type) => JobExecutionAttributes.CreateFromType((Type)type));
 
     public static JobExecutionAttributes GetJobExecutionAttributes(MethodInfo methodInfo) =>
-        Cache.GetOrAdd(methodInfo, (info) => JobExecutionAttributes.CreateFromMethodInfo(info as MethodInfo));
+        Cache.GetOrAdd(methodInfo, (info) => JobExecutionAttributes.CreateFromMethodInfo((MethodInfo)info));
 }
--- a/src/NCronJob/Registry/JobAttributeCache.cs
+++ b/src/NCronJob/Registry/JobAttributeCache.cs
@@ -1,0 +1,14 @@
+using System.Collections.Concurrent;
+using System.Reflection;
+
+namespace NCronJob;
+internal static class JobAttributeCache
+{
+    private static readonly ConcurrentDictionary<object, JobExecutionAttributes> Cache = new();
+
+    public static JobExecutionAttributes GetJobExecutionAttributes(Type jobType) =>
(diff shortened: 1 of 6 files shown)


sample/.editorconfig
```diff
@@ -13,3 +13,4 @@ dotnet_diagnostic.CA1848.severity = none
 dotnet_diagnostic.S2629.severity = none
 dotnet_diagnostic.S2696.severity = none
 dotnet_diagnostic.CA2254.severity = none
+dotnet_diagnostic.RE0001.severity = none
```
linkdotnet, line 16: I guess this can be removed!?
Don't see where you added a Regex.

src/NCronJob/Registry/JobAttributeCache.cs
```diff
@@ -0,0 +1,14 @@
+using System.Collections.Concurrent;
+using System.Reflection;
+
+namespace NCronJob;
+internal static class JobAttributeCache
+{
+    private static readonly ConcurrentDictionary<object, JobExecutionAttributes> Cache = new();
+
+    public static JobExecutionAttributes GetJobExecutionAttributes(Type jobType) =>
+        Cache.GetOrAdd(jobType, (type) => JobExecutionAttributes.CreateFromType(type as Type));
```
linkdotnet, line 10: Do we need the safe cast? If we hard-cast to `Type` we can remove the null-check.

src/NCronJob/Registry/JobExecutionAttributes.cs
```diff
@@ -34,27 +34,49 @@ internal sealed class JobExecutionAttributes
...
+    }
+
+    private JobExecutionAttributes(RetryPolicyAttribute? retryPolicy, SupportsConcurrencyAttribute? concurrencyPolicy)
+    {
+        RetryPolicy = retryPolicy;
+        ConcurrencyPolicy = concurrencyPolicy;
+    }
+
+    // Internal factory methods for cache usage
+    internal static JobExecutionAttributes CreateFromType(Type? jobType)
```
linkdotnet, line 69: This then could be `Type`

linkdotnet:
Some smaller remarks - but otherwise good to merge.

sample/.editorconfig
falvarez1: I didn't add any regex. Had to add that to keep Visual Studio from complaining about the regex on the Cron parameters. It's from the change you made adding the following: `[StringSyntax(StringSyntaxAttribute.Regex)] string cronExpression`

Was that unintentional?

linkdotnet: Oh - yes. Not sure why I thought that should help with regex -.-

Sorry can you please remove the `[StringSyntax(StringSyntaxAttribute.Regex)]` part.

falvarez1: done

src/NCronJob/Registry/JobAttributeCache.cs
falvarez1: hard casting works, done

linkdotnet (approved):

PR #69, opened 2024-05-22
feat: Ability to configure global settings

linkdotnet:
## Motivation

In an advanced scenario, a user might want to control the degree of parallel executed tasks.
For that, we offer a new `GlobalOptions` object.

## Points to consider
Design-wise the new configuration is not available on the *Minimal Job API*. The reason is that it's more likely that a user calls this more than once:
```csharp
builder.Services.AddNCronJob(() => {}, ...);
builder.Services.AddNCronJob(() => {}, ...);
```

But it is less likely that the user calls the "full" version multiple times. Nevertheless this option is global and influences the minimal job API as well.

That said, there is a drawback to the current approach. **If** the user anyway decides to call the "full" `AddNCronJob`multiple times, only the first one is put into the container and is used for the QueueWorker and Executor, **but** we evaluate the `SupportsConcurrency` attribute based on the one provided by the user. So this can lead to a mismatch and runtime issues.

Alternatively we can throw an exception if the user already registered global options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to **NCronJob** will be documented in this file. The project
 
 ## [Unreleased]
 
+### Added
+- Ability to configure the global level of concurrency for all jobs.
+```csharp
+builder.Services.AddNCronJob(
+    jobBuilder => jobBuilder.AddJob<MyJob>(),
+    globalOptions => globalOptions.MaxDegreeOfParallelism = Environment.ProcessorCount * 4);
+```
+
 ## [2.5.0] - 2024-05-21
 
 ### Changed
--- a/docs/advanced/global-concurrency.md
+++ b/docs/advanced/global-concurrency.md
@@ -14,3 +14,14 @@ A simple example: You have only one processor (therefore maximum 4 jobs executed
 So after four minutes the queue is full, and no more jobs will be added to the queue. After the fifth minute, the queue is still full, and no more jobs will be added to the queue. After the sixth minute, the first job is removed from the queue, and the next job is added to the queue. Therefore it can happen that jobs are skipped and not executed.
 
 The same applies to the `SupportsConcurrencyAttribute` discussed in: *[Concurrency Control](../features/concurrency-control.md)*
+
+## Configuration
+**NCronJob** offers the ability to change the default via an optional overload:
+
+```csharp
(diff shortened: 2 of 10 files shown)


falvarez1:
> Alternatively we can throw an exception if the user already registered global options.

I generally prefer when an API handles errors gracefully and corrects mistakes silently instead of throwing exceptions. This approach may also reduce the need for additional error handling code.

falvarez1:
> **If** the user anyway decides to call the "full" `AddNCronJob`multiple times, only the first one is put into the container and is used for the QueueWorker and Executor, **but** we evaluate the `SupportsConcurrency` attribute based on the one provided by the user. So this can lead to a mismatch and runtime issues.

Since Global Options should never change is there something you can do to register it as a singleton? `TryAddSingleton` would prevent overriding on any subsequent calls to `.AddNCronJob()`

linkdotnet:
> > Alternatively we can throw an exception if the user already registered global options.
> 
> I generally prefer when an API handles errors gracefully and corrects mistakes silently instead of throwing exceptions. It makes the dev experience more pleasant. This approach may also reduce the need for additional error handling code.

Will add that

linkdotnet:
> > **If** the user anyway decides to call the "full" `AddNCronJob`multiple times, only the first one is put into the container and is used for the QueueWorker and Executor, **but** we evaluate the `SupportsConcurrency` attribute based on the one provided by the user. So this can lead to a mismatch and runtime issues.
> 
> Since Global Options should never change is there something you can do to register it as a singleton? `TryAddSingleton` would prevent overriding on any subsequent calls to `.AddNCronJob()`

Yes that is my way of thinking - plus adding it to the docs (inline as well as our web page). Not perfect, but hopefully good enough.

linkdotnet:
There is still an edge case, which is failing right now:
```csharp
Services.AddNCronJob(b => b.AddJob<Job1>());
Services.AddNCronJob(b => b.AddJob<Job2>(), t => t.MaxDegreeOfParallelism = 10); // Fails as we already have an instance
```

linkdotnet:
Probably the most straightforward way would be to go in the direction of #37 

That would remove **all** the ambiguity and limit the surface of errors.

falvarez1:
FYI
https://github.com/NCronJob-Dev/NCronJob/commit/4fc5ae0edcf478114e5ef6e886e031fb8110bf78

PR #70, opened 2024-05-22
feat: Support Startup Jobs #43

### LINKED ISSUE #43: Allow "one-time" jobs
Allow the execution of jobs that are only run once at the beginning of the lifecycle. This could be some migration work or other tasks that should run very early on in the lifecycle.

## Example
```csharp
Services.AddNCronJob(options => 
{
  options.AddJob<MyJob>().RunOnce();
});
```

This will lead to a direct execution. `RunOnce` can have two overloads:
```diff
public sealed class JobOptionBuilder
{
+   public ParameterBuilder RunOnce();
+   public ParameterBuilder RunOnce(TimeSpan delay);
+   public ParameterBuilder RunOnce(DateTimeOffset absoluteDay);
}
```

To make that work, the `CronRegistry` has to be extended to distinguish between one-time jobs and CRON jobs.
```diff
internal sealed partial class CronRegistry 
{
+   public IReadOnlyCollection<RegistryEntry> GetAllOneTimeJobs() => oneTimeJobs;
}
```

This then can be picked up by the `CronSchedulder` to execute those once.

falvarez1:
### Description
This PR introduces support for "startup jobs" and "one-time jobs" in the NCronJob library. These enhancements allow for tasks that need to be executed once during the application startup phase or at a specific moment, respectively. Key changes include:

- **Startup Jobs**: Execute once at the beginning of the application lifecycle, suitable for tasks like data migrations or initial data setups.
- **Sample Project**: Provide sample project that demonstrates the new feature.
- **Improved API**: Limit API methods to better guide the developer during setup.

### Related Issue
Closes #43

### Checklist
- [X] My code follows the style guidelines of this project and the AspNetCore coding standards.
- [X] My change requires updates to the documentation.
- [x]  I have updated the documentation accordingly.
- [x] I have added a subsection in the CHANGELOG.md under the appropriate release.
- [ ] I have added, updated, or removed tests as necessary according to my changes.
- [X] All tests passed.


src/NCronJob/Scheduler/QueueWorker.cs
```diff
@@ -57,10 +60,14 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         stopToken.Register(LogCancellationRequestedInJob);
         var runningTasks = new List<Task>();
 
+        await startupJobManager.ProcessStartupJobs(CreateExecutionTask, stopToken);
+
         ScheduleInitialJobs();
 
         try
         {
+            await startupJobManager.WaitForStartupJobsCompletion();
```
linkdotnet, line 69: That is a very good point - migrations and so on should be executed before anything else (even other jobs are started).

The "issue" is that they bypass our other boundaries like retry-policy/concurrency control.

falvarez1: Actually retry will be honored since that's handled by the `IRetryHandler` which happens in the `JobExecutor`. However concurrency management will not. But realistically what are the odds that you would need concurrency management within startup jobs? They won't have to compete for resources with regular jobs (given that they need to finish before any normal jobs start up).

falvarez1: And I guess that since they only run once then you can't (or shouldn't) overlap them like this:
```csharp
builder.Services.AddNCronJob(b =>
    b.AddJob<RunAtStartJob>()
        .RunAtStartup()
        .RunAtStartup());

builder.Services.AddNCronJob(b => b.AddJob<RunAtStartJob>(p =>
    {
        p.RunAtStartup();
        p.RunAtStartup();
    }
));

builder.Services.AddNCronJob(b => b
    .AddJob<RunAtStartJob>(p =>
        p.RunAtStartup())
    .AddJob<RunAtStartJob>(p =>
        p.RunAtStartup())
);
```

falvarez1: I'm dropping support for chaining on to `JobOptionBuilder`, seems too messy and confusing. This will prevent someone from doing this:

```csharp
builder.Services.AddNCronJob(b => b.AddJob<RunAtStartJob>(p =>
    {
        p.RunAtStartup();
        p.RunAtStartup();
    }
));
...
builder.Services.AddNCronJob(b => b
    .AddJob<RunAtStartJob>(p =>
        p.RunAtStartup())
    .AddJob<RunAtStartJob>(p =>
        p.RunAtStartup())
);
```

The API should guide users to set it up in only one way, from `NCronJobOptionBuilder<TJob>` like this:
```csharp
builder.Services.AddNCronJob(b => b.AddJob<RunAtStartJob>().RunAtStartup());
...
builder.Services.AddNCronJob(options =>
{
    options.AddJob<RunAtStartJob>()
        .RunAtStartup()
        .AddNotificationHandler<RunAtStartJobHandler>();
});
```

linkdotnet: That is a very good idea!

src/NCronJob/Configuration/JobOptionBuilder.cs
```diff
@@ -39,6 +40,50 @@ public ParameterBuilder WithCronExpression(string cronExpression, bool? enableSe
...
+        return new ParameterBuilder(this, jobOption);
+    }
+
+
+    /// <summary>
+    /// Configures the job to run once after a specified delay.
+    /// </summary>
+    /// <param name="delay">The delay after which the job should be executed.</param>
+    /// <returns>Returns a <see cref="ParameterBuilder"/> that allows adding parameters to the job.</returns>
+    public ParameterBuilder RunOnce(TimeSpan delay)
```
linkdotnet, line 65: See [#43 ](https://github.com/NCronJob-Dev/NCronJob/issues/43#issuecomment-2125618158)

I somewhat doubt that we really need the delay.
I am trying to come up with a use-case but really can't.

src/NCronJob/Execution/StartupJobManager.cs
```diff
@@ -0,0 +1,33 @@
...
+            await Task.WhenAll(startupTasks);
+        }
+
+        startupJobsCompleted.Set();
+    }
+
+    public Task WaitForStartupJobsCompletion() => startupJobsCompleted.WaitAsync();
+}
+
+internal class AsyncManualResetEvent
```
linkdotnet, line 23: Good idea to have it in itws own class

src/NCronJob/Execution/StartupJobManager.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    private volatile TaskCompletionSource<bool> tcs = new(TaskCreationOptions.RunContinuationsAsynchronously);
```
linkdotnet, line 25: This will only be called be the StartupJobManager, so doesn't have to be `volatile`

src/NCronJob/Configuration/JobOptionBuilder.cs
falvarez1: No worries, I'll remove it. You still want the one that passes absoluteDateTime?

linkdotnet: I would argue that is even less likely than a delay :D
But that is just my current horizon. If a user wants that, we still can add that later on.

src/NCronJob/Scheduler/QueueWorker.cs
falvarez1: This commit enhances the API to guide the dev
https://github.com/NCronJob-Dev/NCronJob/pull/70/commits/8b4525df3d71edad0975590d33dcf0710b885153

src/NCronJob/Configuration/JobOptionBuilder.cs
falvarez1: RunOnce was removed

docs/features/minimal-api.md
```diff
@@ -64,7 +65,7 @@ builder.Services.AddNCronJob([RetryPolicy(retryCount: 3)] (JobExecutionContext c
 }, "0 * * * *");
 ```
 
-## Concurrency-Support
+## Concurrency-Support
```
linkdotnet, line 68: I swear to god I don't know why `Cmd+Space` is a thing on MacBooks leading to this "non-printable whitespace" which clusterf**ks markdown.

src/NCronJob/Registry/JobDefinition.cs
```diff
@@ -15,6 +15,10 @@ internal sealed record JobDefinition(
 
     public CancellationToken CancellationToken { get; set; }
 
+    public bool IsOneTimeJob=> (CronExpression is null && RunAt != null) || IsStartupJob;
+    public DateTimeOffset? RunAt { get; set; }
```
linkdotnet, line 19: Is this still needed?

src/NCronJob/Execution/StartupJobManager.cs
```diff
@@ -0,0 +1,33 @@
+namespace NCronJob;
+internal class StartupJobManager(JobRegistry jobRegistry)
```
linkdotnet, line 2: For reasons of consistency, I would refrain from using primary constructors.

src/NCronJob/Configuration/NCronJobOptionBuilder.cs
```diff
@@ -154,51 +162,183 @@ internal static string GenerateJobName(Delegate jobDelegate)
         var jobHash = jobNameBuilder.ToString().GenerateConsistentShortHash();
         return $"AnonymousJob_{jobHash}";
     }
+
+    internal void RegisterJobs()
+    {
+        foreach (var job in jobs)
+        {
+            Services.AddSingleton(job);
```
linkdotnet, line 170: Shall we use `TryAddSingleton`?

docs/features/startup-jobs.md
```diff
@@ -0,0 +1,72 @@
+# Running Startup Jobs
+
+**NCronJob** allows you to configure jobs to run at application startup. This is useful for tasks that need to be executed immediately when the application starts, such as initial data loading, cleanup tasks, or other setup procedures.
+
+Startup jobs are defined like regular CRON jobs and inherit from `IJob`. The only difference is that they are configured to run at startup.
```
linkdotnet, line 5: Probably we can also mention that **all** startup jobs have to be completed before any other CRON (or instant) job starts.

src/NCronJob/Registry/JobDefinition.cs
falvarez1: Not particularly for this PR, but I added the same to another branch I'm working on for a bug fix related to instant jobs, so I just left it. I'll can remove it to not have to deal with merge issues.

src/NCronJob/Execution/StartupJobManager.cs
falvarez1: Ok, we should add an analyzer rule too.

linkdotnet: Well technically I don't think there is one. I marked `IDE0290` as none.

Not that I am absolutely against it, but I do think we need some shared guidelines and "enforce" them to some extent.

linkdotnet: So I am absolutely open to challenge the current coding standard/guidelines, especially if you feel that they are in your way and hinder productivity.

falvarez1: I like the compactness of primary constructors. The only time they seem to get in the way is when using the source generated `LoggerMessage`.

src/NCronJob/Configuration/NCronJobOptionBuilder.cs
falvarez1: This registration is a replacement for this:
```csharp
foreach (var option in jobOptions)
{
    var cron = option.CronExpression is not null
        ? GetCronExpression(option)
        : null;
    var entry = new JobDefinition(typeof(T), option.Parameter, cron, option.TimeZoneInfo);
--->Services.AddSingleton(entry);
}
```

So I don't think we want to do that otherwise you wont be able to register multiple schedules per Job type.

src/NCronJob/Configuration/NCronJobOptionBuilder.cs
```diff
@@ -154,51 +162,183 @@ internal static string GenerateJobName(Delegate jobDelegate)
...
 
+    /// <inheritdoc />
+    public IStartupStage<TNewJob> AddJob<TNewJob>(Action<JobOptionBuilder>? options = null) where TNewJob : class, IJob =>
+        new NCronJobOptionBuilder(services, settings, jobs).AddJob<TNewJob>(options);
+}
+
+/// <summary>
+/// Defines the contract for a stage in the job lifecycle.
+/// </summary>
+public interface IJobStage
```
linkdotnet, line 281: As this is never returned from any builder-method, does it need this interface in the first place?

linkdotnet: Ahh now I see.

linkdotnet (approved):

PR #71, opened 2024-05-26
refactor: Make types internal as users should only use interface

linkdotnet:
We don't need to make the implementation public


linkdotnet:
I will merge the PR - let me know if this was on purpose, then I will revert it.

falvarez1:
It's ok, making them internal is fine. 

The RunAt and IsOneTimeJob shouldn't be a part of that PR.
I was using those properties in a separate branch to deal with state management and they came along with my cherry picking.

linkdotnet:
Sorry that I deleted them - I thought they were a left-over as in the PR we discussed to delete that line.

PR #72, opened 2024-05-26
refactor: Introduce JobRun

linkdotnet:
I introduced a `JobRun` type that can be used further in the future.

I am not 100% sure whether or not the `JobRunRegistry` is a very good decision as it keeps memory around, which can add up over time. but maybe my concern isn't "valid".


linkdotnet:
The basic idea is that we can easily add new things like `JobState` etc for the dashboard and so on.

PR #82, opened 2024-06-01
refactor: Simplify RemoveChars

linkdotnet:
Small refactoring

### DIFF
--- a/src/NCronJob/Configuration/DynamicJobNameGenerator.cs
+++ b/src/NCronJob/Configuration/DynamicJobNameGenerator.cs
@@ -10,7 +10,7 @@ internal static class DynamicJobNameGenerator
     /// </summary>
     /// <param name="jobDelegate">The delegate to generate a name for</param>
     /// <returns></returns>
-    internal static string GenerateJobName(Delegate jobDelegate)
+    public static string GenerateJobName(Delegate jobDelegate)
     {
         var methodInfo = jobDelegate.GetMethodInfo();
         var jobNameBuilder = new StringBuilder(methodInfo.DeclaringType!.FullName);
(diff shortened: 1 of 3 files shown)


linkdotnet:
I took the liberty to enable C# 13 features for **NCronJob**.  Wanted to get your thoughts on this one @falvarez1

falvarez1 (commented):
The new params collections are pretty nice. Since you made the method public how does that impact the ability to use NCronJob from a .Net 8, C#12 application? Is it binary/source compatible? If this can be used from a .Net 8 app without needing to support C#13 then keep it as-is, otherwise it'll need the corresponding preprocessor directives.

linkdotnet:
> The new params collections are pretty nice. Since you made the method public how does that impact the ability to use NCronJob from a .Net 8, C#12 application? Is it binary/source compatible? If this can be used from a .Net 8 app without needing to support C#13 then keep it as-is, otherwise it'll need the corresponding preprocessor directives.

The class is still `internal`. I try to stick to Microsofts coding guidelines, where they make `class`es `internal` and every function `public` instead of `internal` as it seems redundant.

The `params Span` is a SDK thing. So you need netsdk 9 to compile this, but there is no problem in using in .NET 8.
Otherwise our test suite hopefully wouldn't run!? Well at least that is my way of thinking, but we can easily skip that as this is a microoptimization.

falvarez1 (approved):
gtg

falvarez1:
FYI, was curious about the difference in allocations so I ran a benchmark, seems to be negligible with the previous approach.
image

linkdotnet:
Interesting - really neglecable. While I am still in favor of `Contains`, I am more indifferent about `params ReadOnlySpan<char>`.

PR #85, opened 2024-06-07
Dedicated Queue workers per Job Type

### LINKED ISSUE #34: `CancelledJobIsStillAValidExecution` is not deterministic
The following test is flaky and has to be fixed: `CancelledJobIsStillAValidExecution`.
The problem is that this test seems to fail only when executed with other tests.

EDIT:
The test utilizes a `Task.Delay(200)` to reduce flakiness. The only way to tackle this properly is to introduce some kind of notification layer so that the test has a chance to know when certain operations are done.

### LINKED ISSUE #65: Allow bypass concurrency requirements for instant jobs
Follow-up to #61 and the discussion.

## Motivation
Basically, users might want to be able to trigger a job and be sure that this job gets executed independent of whether or not the concurrency requirement is met.

So if we have a job that allows only one run at a time and there is a job running, we still want to be able to say: "Hey I don't care about those requirements, execute that job anyway!".

falvarez1:
## Pull request description
This PR enhances the core scheduling engine to prevent jobs of different job types from affecting each other. Some benefits are:

- **Improved Scheduling Accuracy**: Scheduled runs occur closer to the actual scheduled time because jobs don't have to compete for the same queue.
- **Isolation of Long-Running Jobs**: Long-running jobs of one type do not affect the scheduling and execution of jobs of other types. This resolves edge cases where concurrent long-running jobs could overload the general queue.

Additionally, this PR introduces job state notifications, job event hooks, and queue notifications for improved observability. While these features may not be immediately utilized, they will be valuable for upcoming planned work aimed at enhancing observability via a dashboard.

### PR meta checklist

- [x] All tests passed.

Closes #65 
Fixes #34


src/NCronJob/Execution/JobExecutor.cs
```diff
@@ -69,11 +69,21 @@ public async Task RunJob(JobRun run, CancellationToken stoppingToken)
...
+        var job = scopedServiceProvider.GetService(definition.Type);
+        if (job != null)
+            return (IJob)job;
 
-    private IJob ResolveJob(IServiceProvider scopedServiceProvider, JobDefinition definition) =>
-        definition.Type == typeof(DynamicJobFactory)
-            ? dynamicJobFactoryRegistry.GetAndDrainJobInstance(scopedServiceProvider, definition)
-            : (IJob)scopedServiceProvider.GetRequiredService(definition.Type);
+        LogUnregisteredJob(definition.Type);
+        job = ActivatorUtilities.CreateInstance(scopedServiceProvider, definition.Type);
```
linkdotnet, line 83: I am still not sure whether or not we want to provide this feature.

This call will only work for parameterless c'tors. But let's keep it for now.

src/NCronJob/Registry/JobState.cs
```diff
@@ -0,0 +1,28 @@
...
+}
+
+internal enum JobStateType
+{
+    Scheduled,
+    Running,
+    Retrying,
+    Completing,
+    Completed,
+    Failed,
```
linkdotnet, line 24: For the sake of consistency to .net itself, how about: `Faulted` as name? It is (currently) internal only anyway.

src/NCronJob/RetryPolicies/RetryHandler.cs
```diff
@@ -57,14 +57,16 @@ public async Task ExecuteAsync(Func<CancellationToken, Task> operation, JobExecu
                 runContext.Attempts++;
                 if (runContext.Attempts > 1)
                 {
+                    runContext.JobRun.NotifyStateChange(JobStateType.Retrying);
```
linkdotnet, line 60: Love it! ❤️

src/NCronJob/Scheduler/TaskScheduler/ConcurrencyTaskScheduler.cs
```diff
@@ -0,0 +1,69 @@
...
+        if (Interlocked.CompareExchange(ref delegatesQueuedOrRunning, maxDegreeOfParallelism, maxDegreeOfParallelism) < maxDegreeOfParallelism)
+        {
+            Interlocked.Increment(ref delegatesQueuedOrRunning);
+            NotifyThreadPoolOfPendingWork();
+        }
+    }
+
+#pragma warning disable CA2008
+    private void NotifyThreadPoolOfPendingWork() =>
+        ThreadPool.UnsafeQueueUserWorkItem(_ =>
```
linkdotnet, line 37: Is there a reason not to use `QueueUserWorkItem`?
Given that this code can be exploitable: https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadpool.unsafequeueuserworkitem?view=net-8.0#system-threading-threadpool-unsafequeueuserworkitem(system-threading-waitcallback-system-object)

src/NCronJob/Scheduler/TaskScheduler/DeterministicTaskScheduler.cs
```diff
@@ -0,0 +1,45 @@
+namespace NCronJob;
+
+/// <summary>
+/// Deterministic and ordered execution: Use when the order of task execution is critical, and tasks may have dependencies that require strict ordering.
+/// </summary>
+internal class DeterministicTaskScheduler : TaskScheduler
+{
+    private readonly LinkedList<Task> tasks = [];
+    private readonly object @lock = new();
```
linkdotnet, line 9: With .NET 9 we can do this:

```csharp
#if NET8_0
    private readonly object @lock = new();
#else
    private readonly Lock @lock = new();
```

src/NCronJob/Scheduler/TaskScheduler/DeterministicTaskScheduler.cs
```diff
@@ -0,0 +1,45 @@
+namespace NCronJob;
+
+/// <summary>
+/// Deterministic and ordered execution: Use when the order of task execution is critical, and tasks may have dependencies that require strict ordering.
+/// </summary>
+internal class DeterministicTaskScheduler : TaskScheduler
```
linkdotnet, line 6: Wouldn't it be easier to use a `ConcurrentQueue` instead of manual locking and using a `LinkedList`?

src/NCronJob/Scheduler/ObservablePriorityQueue.cs
```diff
@@ -0,0 +1,143 @@
...
+    public ObservablePriorityQueue(IComparer<(DateTimeOffset NextRunTime, int Priority)> comparer) : base(comparer)
+    { }
+
+    public void Enqueue([DisallowNull] TElement jobWrapper) =>
+        Enqueue(jobWrapper, (jobWrapper.RunAt ?? DateTimeOffset.UtcNow, (int)jobWrapper.Priority));
+}
+internal class ObservablePriorityQueue<TElement, TPriority> : IEnumerable<TElement>, INotifyCollectionChanged where TPriority : IComparable<TPriority>
+{
+    protected readonly PriorityQueue<TElement, TPriority> PriorityQueue;
+    protected readonly object Lock = new();
```
linkdotnet, line 18: Here we could use the new type as well (`System.Lock`)

src/NCronJob/Scheduler/ObservablePriorityQueue.cs
```diff
@@ -0,0 +1,143 @@
...
+            lock (Lock)
+            {
+                return PriorityQueue.Count;
+            }
+        }
+    }
+
+    private void OnCollectionChanged(NotifyCollectionChangedEventArgs args)
+    {
+        var handler = Interlocked.CompareExchange(ref CollectionChanged, null, null);
```
linkdotnet, line 128: Is this to make the function thread-safe?

Directory.Build.props
```diff
@@ -19,6 +19,7 @@
     </PropertyGroup>
 
     <PropertyGroup>
+        <NoWarn>S2696</NoWarn>
```
linkdotnet, line 22: If we really need to exclude that, I would go for the `.editorconfig`.

linkdotnet (commented):
Really good work! Will try it locally.

src/NCronJob/Execution/JobExecutor.cs
falvarez1: You may be thinking of Activator.CreateInstance()

ActivatorUtilities.CreateInstance() (which takes the IServiceProvider) is used internally in the asp.net core framework and is a part of Microsoft.Extensions.DependencyInjection. If the type has multiple constructors then it will choose the longest one (most params).

https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities.createinstance?view=net-8.0

https://github.com/dotnet/docs/issues/31785

src/NCronJob/Registry/JobState.cs
falvarez1: Yes, of course.

src/NCronJob/Scheduler/TaskScheduler/ConcurrencyTaskScheduler.cs
falvarez1: This was really more of a test to observe what happens during profiling.
The three custom TaskScheduler are still experimental. Currently it's only using the Deterministic one.

Directory.Build.props
falvarez1: Sure. I had to add that as it seemed like it was more of a SonarQube bug than a problem with the code. Or I just couldn't figure out how to change the code and make it happy

src/NCronJob/Scheduler/ObservablePriorityQueue.cs
falvarez1: Yes exactly, it retrieves the current value of CollectionChanged in a thread-safe manner, making sure that the value isn't changed by another thread between the time it's checked and the time it's used.

src/NCronJob/Scheduler/TaskScheduler/DeterministicTaskScheduler.cs
falvarez1: This change required opting into the preview features so I added the following to the csproj file:
```xml
  <PropertyGroup>
    <EnablePreviewFeatures>true</EnablePreviewFeatures>
  </PropertyGroup>
```

linkdotnet: Ahhh I thought they put this from preview to stable. Okay, maybe we should put this up separately or, alternatively, remove this if it doesn't get stable until `net9.0`.

src/NCronJob/Scheduler/ObservablePriorityQueue.cs
linkdotnet: Thanks for clearing that up

src/NCronJob/Execution/JobExecutor.cs
linkdotnet: Ahhh yes I thought of `Activator.CreateInstance`.

Okay - let's roll with that and good that you log a warning.

src/NCronJob/Registry/JobDefinition.cs
```diff
@@ -2,14 +2,15 @@
 
 namespace NCronJob;
 
-internal sealed record JobDefinition(
+internal record JobDefinition(
```
linkdotnet, line 5: Is there a reason that you unsealed the class, just out of curiosity.

src/NCronJob/Registry/JobState.cs
```diff
@@ -0,0 +1,28 @@
+namespace NCronJob;
+
+internal class JobState
+{
+    public JobStateType Type { get; private set; }
+    public DateTime Timestamp { get; private set; }
+    public string Message { get; private set; }
```
linkdotnet, line 7: Maybe I am blind - where is the `Message` used?

src/NCronJob/Registry/JobDefinition.cs
falvarez1: I was using it to define the TElement in the `ObservablePriorityQueue` so it couldn't be sealed. However since I merged your changes where you separated out the JobRun from the JobDefinition, I'm now using JobRun so this can be sealed again.

src/NCronJob/Scheduler/TaskScheduler/DeterministicTaskScheduler.cs
falvarez1: It's up to you. I removed it from `DeterministicTaskScheduler` as I'm now just using ConcurrentQueue for thread management. `ObservablePriorityQueue` is currently the only data structure where we need it.

src/NCronJob/Scheduler/TaskScheduler/DeterministicTaskScheduler.cs
falvarez1: Yes, already made the update.

src/NCronJob/Scheduler/ObservablePriorityQueue.cs
falvarez1: Using the new approach

src/NCronJob/Registry/JobState.cs
falvarez1: The message is currently only set when faulting so it uses the exception's message. My initial thought was that most of the logs and telemetry would reside within the `OnStateChanged` event handler so having the exception message could be valuable.
https://github.com/NCronJob-Dev/NCronJob/blob/9e45fc8fe348f818c2def27e50ac78d040c44ec6/src/NCronJob/Registry/JobRun.cs#L48-L50

src/NCronJob/Scheduler/TaskScheduler/DeterministicTaskScheduler.cs
linkdotnet: Let's keep it - my hope is that it gets stable. We can switch it prior to .net 9 GA.

src/NCronJob/Scheduler/TaskScheduler/ConcurrencyTaskScheduler.cs
linkdotnet: That said, we may want to squash **some** of the commits :D - given this beauty:

> [Hopefully we don't merge that](https://github.com/NCronJob-Dev/NCronJob/pull/85/commits/a53c9b0879bc5287098f148238abee6ab33ebfff)

linkdotnet:
Locally one or more tests are failing:
```no-class
NCronJob.Tests test net9.0 failed with 2 error(s) (13.7s)
    /usr/local/share/dotnet/sdk/9.0.100-preview.4.24267.66/Microsoft.TestPlatform.targets(46,5): error : [xUnit.net 00:00:00.22]     NCronJob.Tests.NCronJobIntegrationTests.InstantJobHasHigherPriorityThanCronJob [FAIL]
    /Users/stevengiesel/repos/NCronJob/tests/NCronJob.Tests/NCronJobIntegrationTests.cs(301): error VSTEST1: 
      NCronJob.Tests.NCronJobIntegrationTests.InstantJobHasHigherPriorityThanCro
      nJob: NCronJob.Tests.NCronJobIntegrationTests.InstantJobHasHigherPriorityT
      hanCronJob() Shouldly.ShouldAssertException : answer
          should be
      "INSTANT"
          but was
      "CRON"
  NCronJob.Tests test net8.0 failed with 2 error(s) (14.7s)
    /usr/local/share/dotnet/sdk/9.0.100-preview.4.24267.66/Microsoft.TestPlatform.targets(46,5): error : [xUnit.net 00:00:00.40]     NCronJob.Tests.NCronJobIntegrationTests.InstantJobHasHigherPriorityThanCronJob [FAIL]
    /Users/stevengiesel/repos/NCronJob/tests/NCronJob.Tests/NCronJobIntegrationTests.cs(301): error VSTEST1: 
      NCronJob.Tests.NCronJobIntegrationTests.InstantJobHasHigherPriorityThanCro
      nJob: NCronJob.Tests.NCronJobIntegrationTests.InstantJobHasHigherPriorityT
      hanCronJob() Shouldly.ShouldAssertException : answer
          should be
      "INSTANT"
          but was
      "Job was canceled"

Test summary: total: 110, failed: 2, succeeded: 108, skipped: 0, duration: 14.7s
```
  NCronJob.Tests test net9.0 failed with 8 error(s) (13.7s)
    /usr/local/share/dotnet/sdk/9.0.100-preview.4.24267.66/Microsoft.TestPlatform.targets(46,5): error : [xUnit.net 00:00:05.14]     NCronJob.Tests.NCronJobIntegrationTests.CanRunSecondPrecisionAndMinutePrecisionJobs [FAIL]
    /Users/stevengiesel/repos/NCronJob/tests/NCronJob.Tests/NCronJobIntegrationTests.cs(154): error VSTEST1: 
      NCronJob.Tests.NCronJobIntegrationTests.CanRunSecondPrecisionAndMinutePrec
      isionJobs: NCronJob.Tests.NCronJobIntegrationTests.CanRunSecondPrecisionAn
      dMinutePrecisionJobs() Shouldly.ShouldAssertException : jobFinished
          should be
      True
          but was
      False
    /usr/local/share/dotnet/sdk/9.0.100-preview.4.24267.66/Microsoft.TestPlatform.targets(46,5): error : [xUnit.net 00:00:05.22]     NCronJob.Tests.NCronJobIntegrationTests.InstantJobHasHigherPriorityThanCronJob [FAIL]
    /Users/stevengiesel/repos/NCronJob/tests/NCronJob.Tests/NCronJobIntegrationTests.cs(301): error VSTEST1: 
      NCronJob.Tests.NCronJobIntegrationTests.InstantJobHasHigherPriorityThanCro
      nJob: NCronJob.Tests.NCronJobIntegrationTests.InstantJobHasHigherPriorityT
      hanCronJob() Shouldly.ShouldAssertException : answer
          should be
      "INSTANT"
          but was
      "Job was canceled"
    /usr/local/share/dotnet/sdk/9.0.100-preview.4.24267.66/Microsoft.TestPlatform.targets(46,5): error : [xUnit.net 00:00:09.26]     NCronJob.Tests.NCronJobRetryTests.CancelledJobIsStillAValidExecution [FAIL]
    /usr/local/share/dotnet/sdk/9.0.100-preview.4.24267.66/Microsoft.TestPlatform.targets(46,5): error : [xUnit.net 00:00:10.27]     NCronJob.Tests.NCronJobIntegrationTests.AdvancingTheWholeTimeShouldHaveTenEntries [FAIL]
    /Users/stevengiesel/repos/NCronJob/tests/NCronJob.Tests/NCronJobRetryTests.cs(115): error VSTEST1: 
      NCronJob.Tests.NCronJobRetryTests.CancelledJobIsStillAValidExecution: NCro
      nJob.Tests.NCronJobRetryTests.CancelledJobIsStillAValidExecution() Shouldl
      y.ShouldAssertException : jobRun.JobExecutionCount
          should be
      1
          but was
      0
    /Users/stevengiesel/repos/NCronJob/tests/NCronJob.Tests/NCronJobIntegrationTests.cs(40): error VSTEST1: 
      NCronJob.Tests.NCronJobIntegrationTests.AdvancingTheWholeTimeShouldHaveTen
      Entries: NCronJob.Tests.NCronJobIntegrationTests.AdvancingTheWholeTimeShou
      ldHaveTenEntries() Shouldly.ShouldAssertException : jobFinished
          should be
      True
          but was
      False

Test summary: total: 110, failed: 4, succeeded: 106, skipped: 0, duration: 13.7s
```

Another run:

File: src/NCronJob/Scheduler/ObservablePriorityQueue.cs [hunk 12]
```diff
@@ -0,0 +1,128 @@
...
+using System.Diagnostics.CodeAnalysis;
+
+namespace NCronJob;
+
+internal class ObservablePriorityQueue<TElement> : ObservablePriorityQueue<TElement, (DateTimeOffset NextRunTime, int Priority)> where TElement : JobRun
+{
+    public ObservablePriorityQueue(IComparer<(DateTimeOffset NextRunTime, int Priority)> comparer) : base(comparer)
+    { }
+
+    public void Enqueue([DisallowNull] TElement jobWrapper) =>
```
linkdotnet, line 12: Is unused

tests/NCronJob.Tests/NCronJobIntegrationTests.cs
```diff
@@ -336,7 +336,14 @@ public async Task AnonymousJobsCanBeExecutedMultipleTimes()
     {
         var fakeTimer = new FakeTimeProvider();
         ServiceCollection.AddSingleton<TimeProvider>(fakeTimer);
-        ServiceCollection.AddNCronJob(n => n.AddJob(async (ChannelWriter<object> writer) => await writer.WriteAsync(true), "* * * * *"));
+        async Task WriteTrueAsync(ChannelWriter<object> writer, CancellationToken ct)
+        {
+            await Task.Delay(10, ct);
```
linkdotnet, line 341: Does it need that delay?

src/NCronJob/Configuration/DynamicJobFactory.cs
```diff
@@ -36,7 +36,8 @@ public DynamicJobFactory(IServiceProvider serviceProvider, Delegate jobAction)
         var param = Expression.Parameter(typeof(object[]), "args");
         var args = method.GetParameters().Select((p, index) =>
             Expression.Convert(Expression.ArrayIndex(param, Expression.Constant(index)), p.ParameterType)).ToArray();
-        var call = Expression.Call(Expression.Constant(jobDelegate.Target), method, args);
+        var instance = method.IsStatic ? null : Expression.Constant(jobDelegate.Target);
```
linkdotnet, line 39: Good catch :D

.editorconfig
```diff
@@ -407,6 +407,8 @@ dotnet_diagnostic.S1118.severity = none        # Add a 'protected' constructor o
 dotnet_diagnostic.S1135.severity = suggestion # https://github.com/atc-net/atc-coding-rules/blob/main/documentation/CodeAnalyzersRules/SonarAnalyzerCSharp/S1135.md
 dotnet_diagnostic.S2583.severity = suggestion # Change this condition so that it does not always evaluate to 'False'.
 dotnet_diagnostic.S3875.severity = none       # Remove this overload of 'operator =='.
+dotnet_diagnostic.S2696.severity = suggestion
```
linkdotnet, line 410: Can be removed again if we remove `ConcurrencyTaskScheduler`

sample/NCronJobSample/Program.cs
```diff
@@ -13,13 +13,22 @@
...
 
     .AddJob<PrintHelloWorldJob>(p =>
-        p.WithCronExpression("*/20 * * * * *", timeZoneInfo: TimeZoneInfo.FindSystemTimeZoneById("Pacific Standard Time")))
+        p.WithCronExpression("*/5 * * * *", timeZoneInfo: TimeZoneInfo.FindSystemTimeZoneById("Pacific Standard Time")))
+
+    .AddJob<PrintHelloWorldJob>(p =>
+        p.WithCronExpression("*/4 * * * * *"))
 
     // Execute the job every 2 minutes
+    .AddJob<DataProcessingJob>(p =>
```
linkdotnet, line 22: It doesn't execute every 2 minutes anymore ;)

src/NCronJob/NCronJobExtensions.cs
```diff
@@ -33,11 +33,14 @@ public static class NCronJobExtensions
         builder.RegisterJobs(); // Complete building the NCronJobOptionBuilder
 
         services.TryAddSingleton(settings);
-        services.AddHostedService<QueueWorker>();
+        services.TryAddSingleton<QueueWorker>();
```
linkdotnet, line 36: Do we need `QueueWorker` somewhere else? That shouldn't be the case

src/NCronJob/Registry/IInstantJobRegistry.cs
```diff
@@ -22,7 +24,7 @@ public interface IInstantJobRegistry
     /// instantJobRegistry.RunInstantJob&lt;MyJob&gt;(new MyParameterObject { Foo = "Bar" });
     /// </code>
     /// </example>
-    void RunInstantJob<TJob>(object? parameter = null, CancellationToken token = default)
+    void RunInstantJob<TJob>(object? parameter = null, bool forceExecution = false, CancellationToken token = default)
```
linkdotnet, line 27: Technically a breaking change.

Another idea would be to have a seperate method:
```csharp
void ForceRunInstantJob<TJob>(object? parameter = null, CancellationToken token = default);
```

src/NCronJob/Registry/JobHistory.cs
```diff
@@ -2,11 +2,26 @@
...
-internal sealed class JobHistory
+internal interface IJobHistory
 {
+    void Add(JobRun jobRun);
+    IReadOnlyCollection<JobRun> GetAll();
+}
+
+internal sealed class JobHistory : IJobHistory
+{
+    //Note: This can grow out of control if not managed properly. JobHistory should ultimately be persisted to a database or something durable.
```
linkdotnet, line 13: Good point - as you said some time ago, we might wanna invest some time into **FASTER**.

We could also have a `FixedSizeBag` with an upper limit.

src/NCronJob/Scheduler/TaskScheduler/TaskFactoryProvider.cs
```diff
@@ -0,0 +1,13 @@
+namespace NCronJob;
+
+internal static class TaskFactoryProvider
```
linkdotnet, line 3: I am much in favor of not commiting those 4 objects (all 3 schedulers and the factory).
If we need them we have the git history.

src/NCronJob/Scheduler/TaskScheduler/TaskFactoryProvider.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    private static readonly TaskFactory TaskFactory = new(
+        CancellationToken.None,
+        TaskCreationOptions.LongRunning | TaskCreationOptions.PreferFairness,
```
linkdotnet, line 7: I would be careful with `LongRunning` here as we "jump" in front of the queue of the TP.
That also means that, potentially, we steal threads for HTTP requests.

If a user offloads tasks with **NcronJob** I guess milliseconds aren't important.

src/NCronJob/Configuration/DynamicJobFactory.cs
linkdotnet: Can you add a test for this?
As far as I understand this should be trivial, just add a `static` in front:
```csharp
instantJobRegistry.RunInstantJob(static async (ChannelWriter<object> writer ) => { ... });
```

tests/NCronJob.Tests/NCronJobIntegrationTests.cs
falvarez1: Without it the Mac test fails sometimes. I think it may have to do with running these Jobs too quickly. The 10ms simulates a tiny bit of work without bogging down the unit tests. It seems like it could have something to do with the FakeTimeProvider

src/NCronJob/Registry/IInstantJobRegistry.cs
falvarez1: good point, didn't want the CancellationToken in the middle, which now is a breaking change.

src/NCronJob/NCronJobExtensions.cs
falvarez1: The `AddNCronJob()` method has potential to be called multiple times by an end user, so we need to make sure we're not creating _multiple_ singletons.

linkdotnet: `AddHostedService` is basically a `TryAddSingleton` under the hood. So I would be surprised if we really have two instances running.

tests/NCronJob.Tests/NCronJobIntegrationTests.cs
linkdotnet: Interesting - recently MS shifted to MacOS ARM64 for the GH runners. We could also try to take the x64 version instead. But that just hides the issue.

src/NCronJob/Scheduler/TaskScheduler/TaskFactoryProvider.cs
falvarez1: Actually, using `LongRunning` doesn't cause tasks to "jump" in front of the queue. The `LongRunning` option indicates to the `TaskScheduler` that the task is expected to run for a long time, potentially prompting it to allocate additional threads to accommodate this. This doesn't prioritize the task over others but helps manage resources more effectively for long-running operations.

If you're referring to the priority queue, it's essentially unordered until the moment of dequeuing or peeking, at which point the job task that meets the JobQueueTupleComparer requirements is processed. This ensures that tasks are executed in the intended order based on their priority without stealing threads meant for other operations like HTTP requests.

src/NCronJob/Scheduler/TaskScheduler/TaskFactoryProvider.cs
falvarez1: I made some changes, kept only the Deterministic scheduler and added a config to ConcurrencySettings to be able to control which one to use.
Also for the `TaskFactoryProvider`, we still need to create a `TaskFactory` with those specific options. This provides cleaner code and SOC.

src/NCronJob/Registry/IInstantJobRegistry.cs
falvarez1: Done. Separated them out into their own methods.

.editorconfig
falvarez1: I removed the `ConcurrencyTaskScheduler` for now. May bring it back in the future if necessary.

src/NCronJob/Configuration/DynamicJobFactory.cs
falvarez1: Done

src/NCronJob/Registry/JobHistory.cs
falvarez1: Right, the idea is that the Write-Ahead Log from FASTER (or equivalent) should ensure data integrity even in the event of a crash. This can also help with rehydrating the data back when the process is restored.

src/NCronJob/Scheduler/TaskScheduler/TaskFactoryProvider.cs
linkdotnet: No, I really was referring to the TaskScheduler of .NET itself and how he handles those "virtual" threads.

Given your expertise, let's roll with it. In my lifetime my need for `LongRunning` was 0 until now.

linkdotnet: My concern is/was that we limit the scalability of the users application.
Or better: We prioritize "our" work (the jobs) over whatever is going on next to it.

src/NCronJob/Registry/JobHistory.cs
linkdotnet: Exactly - probably configurable.

My current thinking is that jobs (by default) shouldn't re-run - but the user explicitly has to ask "us"

src/NCronJob/Registry/IInstantJobRegistry.cs
```diff
@@ -55,101 +65,168 @@ void RunScheduledJob<TJob>(TimeSpan delay, object? parameter = null, Cancellatio
...
+    /// This is a fire-and-forget process, the Job will be run immediately in the background. The contents of <paramref name="parameter" />
+    /// are not serialized and deserialized. It is the reference to the <paramref name="parameter"/>-object that gets passed in.
+    /// </remarks>
+    /// <example>
+    /// Running a job with a parameter:
+    /// <code>
+    /// instantJobRegistry.RunInstantJob&lt;MyJob&gt;(new MyParameterObject { Foo = "Bar" });
+    /// </code>
+    /// </example>
+    void ForceRunInstantJob<TJob>(object? parameter = null, CancellationToken token = default)
```
linkdotnet, line 94: Shall we offer the Minimal API as well or in a follow up PR?

src/NCronJob/Registry/IInstantJobRegistry.cs
```diff
@@ -55,101 +65,168 @@ void RunScheduledJob<TJob>(TimeSpan delay, object? parameter = null, Cancellatio
...
+        where TJob : IJob;
+
+    /// <summary>
+    /// Runs a job that will be executed after the given <paramref name="delay"/>. The job will not be queued into the JobQueue, but executed directly.
+    /// The concurrency settings will be ignored.
+    /// </summary>
+    /// <param name="delay">The delay until the job will be executed.</param>
+    /// <param name="parameter">An optional parameter that is passed down as the <see cref="JobExecutionContext"/> to the job.</param>
+    /// <param name="token">An optional token to cancel the job.</param>
+    void ForceRunInstantJob<TJob>(TimeSpan delay, object? parameter = null, CancellationToken token = default)
```
linkdotnet, line 104: To stick to the naming - that should be `ForceRunScheduledJob`.
Also, here, the Minimal API would be nice (it can be followed up as well).

Also, we would need `ForceRunScheduledJob(DateTimeOffset absolute, ....)` just to have API symmetry.

src/NCronJob/Registry/JobRun.cs
```diff
@@ -1,41 +1,100 @@
...
 
     public static JobRun Create(JobDefinition jobDefinition, object? parameter, CancellationToken token) =>
-        new()
+        new JobRun
         {
             JobRunId = Guid.NewGuid(),
             JobDefinition = jobDefinition,
             Parameter = parameter,
             CancellationToken = token
+        }.Initialize();
```
linkdotnet, line 45: Can't we move this logic to the c'tor? 
The API surface doesn't feel right.

linkdotnet:
We are very close and I just getting nitpicky here ;-)

src/NCronJob/NCronJobExtensions.cs
falvarez1: I see what you mean now. The line right after that says to wire up the QueueWorker as a HostedService. My first attempt was to get access to all of the public dictionaries so that we can surface the Job details to an API or to the tests. Since I broke up the one class into more isolated constructs we don't need this anymore. We can add a separate mechanism (separate PR) to get a snapshot of all the current Jobs, Queues and their state.

src/NCronJob/Registry/IInstantJobRegistry.cs
falvarez1: Done

src/NCronJob/Registry/IInstantJobRegistry.cs
falvarez1: I added it just to complete the API.

src/NCronJob/Registry/JobRun.cs
falvarez1: Done

src/NCronJob/Registry/IInstantJobRegistry.cs
```diff
@@ -151,37 +190,65 @@ public void RunScheduledJob<TJob>(DateTimeOffset startDate, object? parameter =
...
-    
-    /// <inheritdoc />
-    public void ForceRunInstantJob<TJob>(TimeSpan delay, object? parameter = null, CancellationToken token = default)
-        where TJob : IJob
-    {
-        var utcNow = timeProvider.GetUtcNow();
-        RunJob<TJob>(utcNow + delay, parameter, true,token);
+        if (forceExecution)
+        {
+            _ = jobProcessor.ProcessJobAsync(run, token);
```
linkdotnet, line 243: With regard to the UI/Dashboard - will "forced instant jobs" be listed if we bypass the whole queue?

linkdotnet:
LGTM!

Two last things:
 1. Can you add the resolved issues and stuff inside the `CHANGELOG.md` for the release?
 2. (Optional): Move some classes inside folders (except the builder, because I started doing that on my PR).
 I do feel we have a lot of objects and need some organization.

linkdotnet (approved):

falvarez1:
> 2\. (Optional): Move some classes inside folders (except the builder, because I started doing that on my PR).
>     I do feel we have a lot of objects and need some organization.

Yes I feel we need to start moving things around to organize. I was hesitant to do this in this PR because it already has many changes and will make merging other ongoing PRs more difficult. I think we can take care of this in a separate, dedicated refactoring/organize PR.



> 1. Can you add the resolved issues and stuff inside the `CHANGELOG.md` for the release?

Will do.

linkdotnet:
>  I think we can take care of this in a separate, dedicated refactoring/organize PR.

Yes! Feel free to merge the PR (or squash&merge if you want)

src/NCronJob/Registry/IInstantJobRegistry.cs
falvarez1: Good question! I think the dashboard needs two visual areas (maybe more): one that represents all the discoverable jobs, so the jobs registry. The other would be for the Job related queues with the ability to drill down into the instances of each job definition. A forced instant job will not show in a queue because it just bypasses it, but should show in any real-time stream and stored audit for that Job Type. I'll mock something up to help visualize, either in HTML or Figma, before committing to the UX.

falvarez1:
@linkdotnet 

I completely overlooked the fact that a forced job bypasses the queue and this cannot be "scheduled". I think I just blindly implemented the additional API methods to reflect the version that goes through the queue. But there's no way to schedule a forced instant job (atm).

Additionally I'm having second thoughts on the API name `ForceRunInstantJob`. It feels too verbose. The interface itself already establishes that we're working with an "instant job registry". What do you think about renaming `ForceRunInstantJob` to something more simple like `Invoke`, `Trigger` or `ExecuteNow`?

linkdotnet:
Maybe for now we just scratch it (revert commit so we can recover) and tackle it later.

maybe the naming is the sole problem here.
Hanford calls it „Enqueue“ which makes more sense.

There is no concept of bypassing.

I am leaning towards that we wait until customer feedback before we really implement that feature.

falvarez1:
I fixed the feature gap so instant jobs can be delayed and scheduled to run once at the desired time. 
I left the API method names the same and updated a variable name to better reflect its purpose.

PR #88, opened 2024-06-18
fix: Don't use Unsafe version for scheduler

linkdotnet:
We probably don't wanna use the unsafe version for no good reason.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 	<ItemGroup>
-      <PackageReference Include="SonarAnalyzer.CSharp" Version="9.26.0.92422">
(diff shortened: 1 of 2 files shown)


falvarez1:
I used that construct cautiously to maximize performance, fully aware of the potential security risks. However, based on my initial, basic profiling tests, I did not notice any statistically significant performance difference. Given the potential benefits like reduced allocations, better diagnostics, and more control over threading, I think it's worth taking a deeper dive. I'll set aside some time for a more thorough analysis.

We can either use the safer method or just set the `ConcurrencySettings.UseDeterministicTaskScheduler` to false to use the regular OTB TaskScheduler for the time being.

References:
https://github.com/dotnet/runtime/issues/27483

https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadpool.unsafequeueuserworkitem?view=net-8.0

falvarez1 (approved):

linkdotnet:
Then let's park it and set the flag to `false` - that let's us keep the original impl.

I will do this on another PR :D

Issue #59, opened 2024-05-20
Create abstraction around `JobExecutionContext` to hide some implemenation details


linkdotnet:
In #53 by @falvarez1 the `JobExecutionContext` has some internal details we don't want to expose. That includes the whole construction of the object itself.

To keep the current contract stable and let users be able to test their code, we had to add a ctor that mimics the old behavior.
That isn't a very nice way. The much cleaner approach is to create a `IJobExecutionContext` interface with the public stuff and make `JobExecutionContext` internal.

## ToDo
- [ ] Create `IJobExecutionContext` interface with the necessary stuff and make `JobExecutionContext` internal
- [ ] Update the docs
- [ ] Create an entry for a migration

linkdotnet:
Done on `v3`

Issue #44, opened 2024-05-07
Remove `enableSecondPrecision`


linkdotnet:
The `enableSecondPrecision` flag is no longer needed since we automatically determine whether or not we want that precision.

So, there is no benefit for the user. While this is a major breaking change, I am still inclined to say it is fine to do in `v2`.

kunalshokeen051:
Hi, can i work on this issue.

linkdotnet:
Hey @kunalshokeen051  - thanks for offering your help. Sure thing! 
If you have any questions, feel free to let us know.

By the way @falvarez1  - do you see any reason against removing the flag and releasing it in `v2`?

falvarez1:
No problems. After adding the automatic detection the only reason for keeping it is backwards compatibility. We just need to document as a breaking change.

linkdotnet:
Yeah, that said, @kunalshokeen051 feel free to make the changes.
You might want to add an entry about the migration to the docs. Basically telling folks that they don't need that entry as it is automatically inferred.

linkdotnet:
Given that there was no response for 2 weeks I will unassign the ticket so anyone who wants to help, can do so.

linkdotnet:
Done on `v3`

PR #90, opened 2024-07-21
v3 branch

linkdotnet:
This is the ongoing branch for `v3`.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,7 +6,7 @@
 	"image": "mcr.microsoft.com/dotnet/sdk:9.0-preview",
 	"features": {
 		"ghcr.io/devcontainers/features/dotnet:2": {
-			"version": "latest"
+			"version": "8.0"
 		}
 	},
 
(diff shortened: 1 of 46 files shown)


linkdotnet:
Fixes #44 
Fixes #59

falvarez1:
@linkdotnet what do you think about converting `Task` to `ValueTask` for `IJob.RunAsync`? Since this already creates breaking changes to IJob if changing to ValueTask improves performance then it can be included. Of course, I would need to run some benchmarks before making any changes.

falvarez1 (approved):

linkdotnet:
> @linkdotnet what do you think about converting `Task` to `ValueTask` for `IJob.RunAsync`? Since this already creates breaking changes to IJob if changing to ValueTask improves performance then it can be included. Of course, I would need to run some benchmarks before making any changes.

That is a good idea - I didn't have very much time lately but those smaller things seemed somewhat in reach without investing hours in one session (what one would need for bigger features).

Coming back - `ValueTask` might be a good choice here as well. Absolutely.

linkdotnet:
I will create a ticket for that.

PR #295, opened 2025-06-21
Minor fixups to devcontainer

nulltoken:
### PR meta checklist

  - [ ] All tests passed.


.devcontainer/post-install.sh
```diff
@@ -7,4 +7,4 @@ apt update
 apt install -y mkdocs
 
 # Trust dotnet developer certs
-dotnet dev-certs https --check --trust
\ No newline at end of file
+dotnet dev-certs https --check --trust
```
nulltoken, line 10: @linkdotnet FWIW, This line does work for me and err with "No valid certificates found"

Is it any required? What's the goal of it?

linkdotnet: If you wanna serve the samples (as they have https) the dotnet dev-certs have to be installed.
So strictly - no not necessary, but a nice goody.

.gitattributes
```diff
@@ -1 +1,2 @@
 * text=auto
+.devcontainer/*.sh text eol=lf
```
linkdotnet, line 2: Isn't `lf` the default?

linkdotnet (approved):

.gitattributes
nulltoken: "auto" makes the checkout/staging compatible with the underlying platform (thanks to the clean/smudge mechanism), while always storing files in git with lf line endings.

For instance, on Windows, with "auto" configuration, files are stored within git with lf line endings, but checking them out store them with crlf line endings on the disk.

"auto" is the sensible choice for 99.9% setups.

The added line overrides this settings by special casing the shell scripts and requesting git to always have them bear lf line endings on disk, independently of the underlying platform.

